PR #8, opened 2021-08-06
Added a linter target with golangci-lint

arjunrn:
[golangci-lint](https://golangci-lint.run/) is a collection of linters which can be configured through a single configuration. Added a make target to download the golangci-lint binary and store it locally. The script can also check the local version and download a newer version if required.

Also included are formatting changes for imports.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,10 @@ verify: verify-scripts
 update-scripts:
 	hack/update-apigen.sh
 .PHONY: update-scripts
+
+ensure-golangci-lint:
+	hack/ensure-golangci-lint.sh
+.PHONY: ensure-golangci-lint
+
+lint: ensure-golangci-lint
+	bin/golangci-lint -c .golangci.yaml run
--- a/hack/ensure-golangci-lint.sh
+++ b/hack/ensure-golangci-lint.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+set -euo pipefail
+
+VERSION=1.41.1
+GOOS=$(go env GOOS)
+TARNAME=golangci-lint-$VERSION-$GOOS-amd64.tar.gz
+URL=https://github.com/golangci/golangci-lint/releases/download/v$VERSION/$TARNAME
+GOLANGCI_LINT=bin/golangci-lint
+
+case $GOOS in
+    linux)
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -19,3 +19,10 @@ verify: verify-scripts
 update-scripts:
 	hack/update-apigen.sh
 .PHONY: update-scripts
+
+ensure-golangci-lint:
```
tjungblu, line 23: just one nit: can we make this part of the verify-scripts target as well? I would not like to run multiple linting and verification steps

hack/ensure-golangci-lint.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+set -euo pipefail
+
+VERSION=1.41.1
+GOOS=$(go env GOOS)
+TARNAME=golangci-lint-$VERSION-$GOOS-amd64.tar.gz
+URL=https://github.com/golangci/golangci-lint/releases/download/v$VERSION/$TARNAME
+GOLANGCI_LINT=bin/golangci-lint
+
+case $GOOS in
```
tjungblu, line 10: are you sure the checksums are correct? I didn't find any of those two hashes in the https://github.com/golangci/golangci-lint/releases/download/v1.41.1/golangci-lint-1.41.1-checksums.txt

hack/ensure-golangci-lint.sh
```diff
@@ -0,0 +1,45 @@
...
+    linux)
+        CHECKSUM=2ad1b2313f89d8ecb35f2e4d2463cf5bc1e057518e4c523565f09416046c21f7
+        ;;
+    darwin)
+        CHECKSUM=458f15c43f72b0bd905ff165c42cf474e5510f91f4a10ae025e42e22fe8b578f
```
tjungblu, line 15: anybody has an m1 on our team? I think for mac we need to support ARM+AMD64 somehow

tjungblu (approved):

hack/ensure-golangci-lint.sh
arjunrn: Those checksums are for the tarballs. But I'm verifying the checkum of the binary to simplify redownloading the binary when we update the version.

hack/ensure-golangci-lint.sh
arjunrn: I'm not sure but this will have to be done at some point. For now just `amd64` support is good enough.

PR #9, opened 2021-08-06
Implemented input directory traversal

arjunrn:
The `FileWalker` type implements a `Traverse()` method which walks the directory tree of the input and then calls the `Outputter` to write the processed output.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -1,12 +1,55 @@
 package cli
 
-import "github.com/openshift/must-gather-clean/pkg/schema"
+import (
+	"errors"
+	"fmt"
+
+	"github.com/openshift/must-gather-clean/pkg/obfuscator"
+	"github.com/openshift/must-gather-clean/pkg/omitter"
+	"github.com/openshift/must-gather-clean/pkg/output"
+	"github.com/openshift/must-gather-clean/pkg/schema"
+	"github.com/openshift/must-gather-clean/pkg/traversal"
+)
 
 func Run(configPath string, inputPath string, outputPath string) error {
 
-	_, err := schema.ReadConfigFromPath(configPath)
+	config, err := schema.ReadConfigFromPath(configPath)
+	if err != nil {
+		return fmt.Errorf("failed to read config at %s: %w", configPath, err)
+	}
+
+	var obfuscators []obfuscator.Obfuscator
+	for _, o := range config.Config.Obfuscate {
+		switch o.Type {
+		case schema.ObfuscateTypeKeywords:
+			obfuscators = append(obfuscators, obfuscator.NewKeywordsObfuscator(o.Replacement))
+		}
+	}
+
+	if len(obfuscators) < 1 {
+		return errors.New("no obfuscator specified in config")
+	}
+
+	var omitters []omitter.Omitter
+	for _, o := range config.Config.Omit {
+		switch om := o.(type) {
+		case schema.FileOmission:
+			omitters = append(omitters, omitter.NewFilenamePatternOmitter(om.Pattern))
+		}
+	}
+
+	writer, err := output.NewFSWriter(outputPath)
 	if err != nil {
 		return err
+	}
+	walker, err := traversal.NewFileWalker(inputPath, writer, obfuscators, omitters)
+	if err != nil {
+		return err
+	}
+
+	err = walker.Traverse()
+	if err != nil {
+		return fmt.Errorf("failed to generate obfuscated output: %w", err)
 	}
 
 	return nil
--- a/pkg/obfuscator/obfuscator.go
+++ b/pkg/obfuscator/obfuscator.go
@@ -1,0 +1,11 @@
+package obfuscator
+
+// Obfuscator is the interface which all obfuscators should implement
+type Obfuscator interface {
+	// FileName takes a filename as input and return the obfuscated name
+	FileName(string) string
+	// Contents takes string as input and return the obfuscated string
+	Contents(string) string
+	// Report returns a mapping of strings which were replaced by this obfuscator
+	Report() map[string]string
+}
--- a/pkg/omitter/fileomitter.go
+++ b/pkg/omitter/fileomitter.go
@@ -1,0 +1,20 @@
+package omitter
+
+import "path/filepath"
+
+type fileOmitter struct {
+	filePattern string
+}
+
+func (f *fileOmitter) File(filename, _ string) (bool, error) {
+	return filepath.Match(f.filePattern, filename)
+}
+
+func (f *fileOmitter) Contents(_ string) (bool, error) {
+	return false, nil
+}
+
+// NewFilenamePatternOmitter return an omitter which omits files based on a globbing pattern.
+func NewFilenamePatternOmitter(pattern string) Omitter {
+	return &fileOmitter{filePattern: pattern}
+}
--- a/pkg/output/output.go
+++ b/pkg/output/output.go
@@ -1,0 +1,76 @@
+package output
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+// Closer is used to close a file when done
+type Closer func() error
+
+// Outputter is an interface for any object which can write the processed output.
+type Outputter interface {
+	// Writer returns a io.StringWriter which can used to write to the output. The caller must ensure that
+	// the parent directory already exists and the close function is called after the caller is done.
+	Writer(parent string, name string, permissions os.FileMode) (Closer, io.StringWriter, error)
+}
+
+type fsWriter struct {
+	outputDir string
+}
+
+func (f *fsWriter) Writer(parentPath, name string, permissions os.FileMode) (Closer, io.StringWriter, error) {
+	filePath := filepath.Join(f.outputDir, parentPath, name)
+	err := os.MkdirAll(filepath.Join(f.outputDir, parentPath), 0755)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create directory %s: %w", parentPath, err)
+	}
+	_, err = os.Stat(filePath)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, nil, fmt.Errorf("failed to determine if %s already exists: %w", filePath, err)
+	}
+	if err == nil {
+		return nil, nil, fmt.Errorf("file %s already exists", filePath)
+	}
+	file, err := os.OpenFile(filePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, permissions)
+	if err != nil {
+		return nil, nil, err
+	}
+	writer := bufio.NewWriter(file)
+	return func() error {
+		if err := writer.Flush(); err != nil {
+			return err
+		}
+		return file.Close()
+	}, writer, nil
+}
+
+// NewFSWriter returns an Outputter which writes the files and directories to a specified location.
+func NewFSWriter(path string) (Outputter, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return nil, err
+		} else {
+			err := os.Mkdir(path, 0700)
+			if err != nil {
+				return nil, fmt.Errorf("failed to create output directory %s: %w", path, err)
+			}
+		}
+	} else {
+		if !info.IsDir() {
+			return nil, fmt.Errorf("output destination must be a directory: %s", path)
+		}
+		entries, err := os.ReadDir(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get contents of output directory %s: %w", path, err)
+		}
+		if len(entries) != 0 {
+			return nil, fmt.Errorf("output directory %s is not empty", path)
+		}
+	}
+	return &fsWriter{outputDir: path}, nil
+}
--- a/pkg/output/output_test.go
+++ b/pkg/output/output_test.go
@@ -1,0 +1,129 @@
+package output
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFSWriter(t *testing.T) {
+	for _, tc := range []struct {
+		name         string
+		writeActions []struct {
+			parent    string
+			name      string
+			contents  []string
+			expectErr string
+		}
+		expectedFiles []struct {
+			path     string
+			contents string
+		}
+	}{
+		{
+			name: "simple-operations",
+			writeActions: []struct {
+				parent    string
+				name      string
+				contents  []string
+				expectErr string
+			}{
+				{parent: "abc/def/ghi", name: "jkw", contents: []string{"file-contents-1", "file-contents-2"}},
+			},
+			expectedFiles: []struct {
+				path     string
+				contents string
+			}{
+				{"abc/def/ghi/jkw", "file-contents-1\nfile-contents-2\n"},
+			},
+		},
+		{
+			name: "invalid write",
+			writeActions: []struct {
+				parent    string
+				name      string
+				contents  []string
+				expectErr string
+			}{
+				{parent: "abc/def/ghi", name: "jkw", contents: []string{"file-contents-1", "file-contents-2"}},
+				{parent: "abc/def/ghi/jkw", name: "invalidfile", contents: []string{"file-contents-1", "file-contents-2"}, expectErr: "jkw: not a directory"},
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			tempdir, err := os.MkdirTemp(os.TempDir(), "test-*")
+			require.NoError(t, err)
+			defer os.RemoveAll(tempdir)
+			writer, err := NewFSWriter(tempdir)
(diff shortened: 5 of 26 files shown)


pkg/output/output.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+)
+
+// Outputter is an interface for any object which can write the processed output.
+type Outputter interface {
+	// Write is used to write content of a file. The caller must ensure that the parent directory already exists.
+	Write(parent string, name string, contents string, permissions os.FileMode) error
```
tjungblu, line 13: can we make this based on line-by-line processing (also on reading)? I'm really afraid of blowing up peoples memory with log files here.

tjungblu (changes requested):

pkg/obfuscator/obfuscator.go
```diff
@@ -0,0 +1,11 @@
+package obfuscator
+
+// Obfuscator is the interface which all obfuscators should implement
+type Obfuscator interface {
+	// FileName takes a filename as input and return the obfuscated name
+	FileName(string) string
+	// Contents takes a file contents as input and return the obfuscated content
+	Contents(string) string
```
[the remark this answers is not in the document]
arjunrn: Not necessarily. The replacements for file contents could be different to what replacements are done in filenames, since file names may not permit some characters while almost everything is permitted in the file contents. At some point filename could also return an error if the replacement rule is not valid for filenames.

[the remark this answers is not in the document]
tjungblu: then we should reflect that as part of the schema as well, otherwise it will be confusing that the same rule is used for filenames and contents

pkg/omitter/fileomitter.go
```diff
@@ -0,0 +1,16 @@
...
+	filePattern string
+}
+
+func (f *fileOmitter) Omit(filename, _, _ string) (bool, error) {
+	return filepath.Match(f.filePattern, filename)
+}
+
+// NewFilenamePatternOmitter return an omitter which omits files based on a globbing pattern.
+func NewFilenamePatternOmitter(pattern string) Omitter {
+	return &fileOmitter{filePattern: pattern}
```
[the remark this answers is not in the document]
arjunrn: How do you mean? Because if the regex is compiled then it doesn't return an error when it is used.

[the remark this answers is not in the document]
tjungblu: where do you want to validate that the user supplied regex is indeed valid?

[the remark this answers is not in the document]
arjunrn: In the `New` method.

pkg/cli/cli.go
```diff
@@ -1,13 +1,56 @@
...
+
+	var omitters []omitter.Omitter
+	for _, o := range config.Config.Omit {
+		switch om := o.(type) {
+		case schema.FileOmission:
+			omitters = append(omitters, omitter.NewFilenamePatternOmitter(om.Pattern))
+		}
+	}
+
+	if len(obfuscators)+len(omitters) < 1 {
```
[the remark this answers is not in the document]
arjunrn: Could you add the validation in the JSON schema?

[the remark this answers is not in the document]
tjungblu: I can, but it's fairly easy by just supplying the minItems property: https://json-schema.org/understanding-json-schema/reference/array.html#length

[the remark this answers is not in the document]
arjunrn: But the `obfuscators` and `omitters` are not in a single array. They are 2 separate properties and the schema needs to specify that either one of them should be present. It's not as simple as specifying minimum length.

pkg/obfuscator/obfuscator.go
```diff
@@ -0,0 +1,11 @@
...
+	// Legend returns a mapping of strings which were replaced by this obfuscator
+	Legend() map[string]string
```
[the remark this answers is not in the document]
arjunrn: Changed it to `Report` because that's what it will be eventually. A report of which words were replace, where and how many times.

pkg/output/output.go
arjunrn: I've changed it to an line-by-line processor.

pkg/cli/cli.go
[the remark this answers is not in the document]
tjungblu: fair enough, I'm just wondering whether it makes sense. There should always be a at least one obfuscator, and there might not be any omissions. The other way seems a bit less likely. Or what do you think?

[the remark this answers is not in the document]
arjunrn: I think the current implementation where we error when there is not obfuscator or omitter is general enough. Why lock down how users use this when there's nothing to be gained.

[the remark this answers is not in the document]
tjungblu: I don't want to bikeshed this any further, but I do assume people want to use the obfuscation (rather than the omission) and when they leave the obfuscation out it's an error and when they leave out the omission it's an acceptable choice.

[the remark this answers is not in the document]
arjunrn: Done

pkg/obfuscator/obfuscator.go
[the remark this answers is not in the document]
arjunrn: Updated the schema.

pkg/cli/cli.go
```diff
@@ -1,13 +1,56 @@
...
+	var obfuscators []obfuscator.Obfuscator
+	for _, o := range config.Config.Obfuscate {
+		switch o.Type {
+		case schema.ObfuscateTypeKeywords:
+			obfuscators = append(obfuscators, obfuscator.NewKeywordsObfuscator(o.Replacement))
+		}
+	}
+
+	if len(obfuscators) < 1 {
+		return errors.New("no omitter specified in schema")
```
tjungblu, line 30: obfuscator or omitter?

arjunrn: Whoops, fixed.

pkg/schema/schema.json
```diff
@@ -19,16 +19,19 @@
...
                                 "type": "MAC",
-                                "replacementType": "Random"
+                                "replacementType": "Random",
+                                "target": "FileContents"
                             }
                         ]
                     ],
                     "title": "Obfuscation Schema",
                     "type": "array",
+                    "minItems": 1,
```
tjungblu, line 34: do you mind adding a small test file in pkg/schema/testfiles/malformed? to ensure that works? I think with that you can also skip the `if len(obfuscators) < 1 {` as the validation during the config read would catch it

pkg/traversal/testfiles/test1/contents.yaml
```diff
@@ -0,0 +1,6 @@
+files:
+  "pods/manifests.yaml": "podkey:podcontents\n"
+  "nodes/test.yaml": "nodekey:nodecontents\n"
+directories:
+  "pods":
+  "nodes":
```
tjungblu, line 6: nit: newline

pkg/traversal/testfiles/test1/mg/nodes/test.yaml
```diff
@@ -0,0 +1 @@
+nodekey:nodecontents
```
tjungblu, line 1: nit: newline

pkg/traversal/testfiles/test1/mg/pods/manifests.yaml
```diff
@@ -0,0 +1 @@
+podkey:podcontents
```
tjungblu, line 1: nit: newline

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,133 @@
...
+			}
+			// If the file should be omitted then stop processing.
+			if omit {
+				continue
+			}
+			return func() error {
+				writeCloser, writer, err := w.writer.Writer(outputDirName, e.Name(), fileInfo.Mode())
+				defer func() {
+					if err := writeCloser(); err != nil {
+						fmt.Printf("failed to successfully write file %s: %v", path.Join(outputDirName, e.Name()), err)
```
tjungblu, line 106: can you please use `filepath` instead of path? it's known to not join properly on windows for example (due to the backslashes and escaping).

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,133 @@
...
+				return err
+			}
+			childDirOutput := path.Join(outputDirName, name)
+			childDirInput := path.Join(inputDirName, e.Name())
+			err = w.processDir(childDirInput, childDirOutput)
+			if err != nil {
+				return err
+			}
+		} else {
+			leafPath := path.Join(inputDirName, e.Name())
```
tjungblu, line 78: filepath vs. path here too. EDIT: okay well there are a couple of places in this file that have path

pkg/traversal/traversal.go
tjungblu: just some background reading: https://stackoverflow.com/questions/39181790/whats-the-difference-between-path-path-filepath-packages-in-go

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,133 @@
...
+
+			var omit bool
+			// verify if the file should be omitted
+			for _, o := range w.omitters {
+				omit, err = o.File(e.Name(), leafPath)
```
tjungblu, line 83: is `e.Name()` the full path from the must-gather root? I believe it's only the filename isn't it?

tjungblu: which is important for globbing, think of scenarios like 
`release-4.10/ingress_controllers/*/haproxy.*`

pkg/obfuscator/obfuscator.go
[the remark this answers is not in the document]
tjungblu: looks good, I think you need to update the comments to also reflect that it's line based

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,133 @@
...
+				if err != nil {
+					return fmt.Errorf("failed to determine if %s should be omitted based on name: %w", leafPath, err)
+				}
+				if omit {
+					break
+				}
+				omit, err = o.Contents(leafPath)
```
tjungblu, line 90: that's the path, not the content - or shall we do this in a follow-up PR?

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,133 @@
...
+		return err
+	}
+	for _, e := range entries {
+		fileInfo, err := e.Info()
+		if err != nil {
+			return err
+		}
+
+		// If a file is a symlink then ignore it
+		if fileInfo.Mode()&os.ModeSymlink != 0 {
```
tjungblu, line 56: good idea, can we add a test for that?

pkg/output/output.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Closer is used to close a file when done
+type Closer func() error
+
+// Outputter is an interface for any object which can write the processed output.
+type Outputter interface {
+	// Writer returns a io.StringWriter which can used to write to the output. The caller must ensure that
+	// the parent directory already exists and the close function is called after the caller is done.
+	Writer(parent string, name string, permissions os.FileMode) (Closer, io.StringWriter, error)
+	// MakeDir creates a directory. The caller must ensure that the parent already exists.
```
tjungblu, line 19: do we need to have an extra call for mkdir? when we attempt to write a file, can we not just `os.MkdirAll`?

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,133 @@
...
+				}()
+				if err != nil {
+					return err
+				}
+				file, err := os.Open(leafPath)
+				if err != nil {
+					return err
+				}
+				scanner := bufio.NewScanner(file)
+				scanner.Split(bufio.ScanLines)
```
tjungblu, line 117: isn't ScanLines the default anyway?

pkg/traversal/traversal.go
```diff
@@ -0,0 +1,133 @@
...
+		// If a file is a symlink then ignore it
+		if fileInfo.Mode()&os.ModeSymlink != 0 {
+			continue
+		}
+
+		if e.IsDir() {
+			name := e.Name()
+			// obfuscate the directory name
+			for _, o := range w.obfuscators {
+				name = o.FileName(name)
```
tjungblu, line 64: now that's confusing: 

```
if e.IsDir() {
   name := o.FileName(e.Name())
}
```

I think we don't have to obfuscate the directory names here. It's enough to do it with the files

pkg/output/output_test.go
```diff
@@ -0,0 +1,170 @@
...
+				contents, err := ioutil.ReadFile(fPath)
+				require.NoError(t, err)
+				require.Equal(t, f.contents, string(contents))
+			}
+		})
+	}
+}
+
+func TestFSWriterNonEmptyDir(t *testing.T) {
+	testDir, err := os.MkdirTemp(os.TempDir(), "test-dir-*")
```
tjungblu, line 140: you can also use `ioutil.TempDir("", "test-dir")`. I think it would be also useful to add an indicator of your test name in the file, so when something fails you can later refer back to the file and debug if possible.

Please also try to clean the directories that you create (even when temp), I found myself deleting a couple of gigs the other day from my own project test files after a year of development.

tjungblu (changes requested):

pkg/schema/schema.json
arjunrn: Validation for `minItems` has not been implemented in the [library](https://github.com/atombender/go-jsonschema#validation)

tjungblu: crap. I'll add it over there, let's leave it here then.

pkg/traversal/traversal.go
arjunrn: Which is why the fullpath is also passed through `leafPath`.

pkg/traversal/traversal.go
arjunrn: We pass the full path here and expect the omitter to implement whatever it needs to read the file.

pkg/traversal/traversal.go
arjunrn: Already exists.

pkg/output/output.go
arjunrn: I'm trying to preserve the permissions of the directories and this requires that we don't use `MkdirAll`.

pkg/traversal/traversal.go
arjunrn: What if the directory name contains a hostname or IP address?

pkg/output/output_test.go
arjunrn: Error messages will have the full path of the failing directory.

pkg/output/output.go
tjungblu: what kind of special permissions do you expect subfolders to have in a must-gather?

pkg/traversal/traversal.go
tjungblu: does that happen? If yes we should either change the `FileName` method to `FilePath` as it can also match directories, or only match on file names

pkg/output/output.go
arjunrn: Also I'm trying to avoid the situation where obfuscation causes directories to be overwritten or merged.

pkg/traversal/traversal.go
arjunrn: Since it's only obfuscating the name(either directory or file) I don't see how changing it to `FilePath` would be more correct.

pkg/schema/schema.json
tjungblu: added it here: https://github.com/atombender/go-jsonschema/pull/30

pkg/output/output.go
tjungblu: I would just say that we don't ever change the directory structure (aside from omissions) and that means that we do not try to obfuscate it. I'm 99% certain that there won't be anything to scrub in the directory names anyway

pkg/traversal/traversal.go
tjungblu: then let's keep the "FileName" method and not obfuscate the directories

pkg/output/output_test.go
tjungblu: how come? they are never included in your require statements

pkg/traversal/traversal.go
tjungblu: where? I don't see a symbolic link in the change list

pkg/output/output_test.go
arjunrn: If any file operations fail then the error at the level will have the path of the file. But if we are cleaning up then all of this is moot anyway because any files in the failing tests will be deleted anyway.

pkg/traversal/traversal.go
arjunrn: 👍🏼

pkg/traversal/traversal.go
arjunrn: [Here](https://github.com/openshift/must-gather-clean/blob/80e178188f7e47b6220a256e6078f220f5f836fb/pkg/traversal/testfiles/test1/mg/nodes/another.yaml)

tjungblu: :+1: the diff didn't show it's a symlink, thx!

PR #12, opened 2021-08-10
add support to obfuscate mac addresses

tjungblu:


pkg/obfuscator/mac_address.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	return s
+}
+
+func (m *macAddressObfuscator) Report() map[string]string {
+	return m.ReplacementReporter.Report()
+}
+
+func NewMacAddressObfuscator() (Obfuscator, error) {
+	regex, err := regexp.Compile(`(?:[0-9a-fA-F]([:-])?){12}`)
```
arjunrn, line 34: `MustCompile()` is a better fit for this because the Mac Address regexp is part of the code.

pkg/obfuscator/replacement_reporter.go
```diff
@@ -0,0 +1,29 @@
...
+
+type SimpleReporter struct {
+	mapping map[string]string
+}
+
+func (s *SimpleReporter) Report() map[string]string {
+	return s.mapping
+}
+
+func (s *SimpleReporter) UpsertReplacement(original string, replacement string) {
```
arjunrn, line 23: Is there actually a use-case for upserting? Because I think if it happens then we've done something wrong.

pkg/obfuscator/mac_address_test.go
```diff
@@ -0,0 +1,50 @@
...
+		{name: "uppercase-colon", input: "69:80:6F:E6:7C:05", expectedOutput: StaticMacReplacement},
+		{name: "lowercase-dash", input: "eb-a1-2a-b2-09-bf", expectedOutput: StaticMacReplacement},
+		{name: "lowercase-colon", input: "eb:a1:2a:b2:09:bf", expectedOutput: StaticMacReplacement},
+		{name: "multi-colon", input: "eb:a1:2a:b2:09:bf eb:a1:2a:b2:09:bf", expectedOutput: StaticMacReplacement + " " + StaticMacReplacement},
+		{name: "multi-colon-dash", input: "16-7C-44-26-24-14 BF:51:A4:1B:7D:0B", expectedOutput: StaticMacReplacement + " " + StaticMacReplacement},
+		{name: "mac surrounded", input: "mac 52:df:20:08:6c:ff caused some trouble", expectedOutput: fmt.Sprintf("mac %s caused some trouble", StaticMacReplacement)},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			o, err := NewMacAddressObfuscator()
+			assert.Nil(t, err)
```
arjunrn, line 46: `require` functions are a much better fit for tests IMO.

pkg/obfuscator/replacement_reporter.go
tjungblu: depends on the replacementType ultimately, I assume we have to sort this out when we do the consistent replacement algorithm.

pkg/obfuscator/mac_address_test.go
tjungblu: for me `assert` is more useful because it shows all errors happening in context and sequence, in this case it doesn't really matter. The error went away due to the mustCompile already :)

pkg/obfuscator/replacement_reporter.go
arjunrn: Even with consistent replacement I don't think we would want to update any existing keys right?

pkg/obfuscator/mac_address_test.go
arjunrn: But after one error is it useful to keep going?

pkg/obfuscator/replacement_reporter.go
tjungblu: who should control that behaviour? is that the "replacer" (which we don't have yet) that ensures it always maps to the same output or the reporter that keeps track of what was replaced? 

I'll change that for the reporting however for additional safety, even though for me that would be a leaking abstraction from the replacer.

pkg/obfuscator/mac_address_test.go
tjungblu: in this case no, because there's nothing else to assert afterwards anyway. Take the testcase above "TestMacReplacementManyMatchLine", it would be useful to know (for me at least) that even though the Contents output wasn't properly replaced the reporting may have worked correctly. 

I'm always looking at it from a CI and debuggability perspective, the easier I can track down what's wrong from the CI logs the better. It becomes a bit harder if it stops at the first assertion already.

pkg/obfuscator/replacement_reporter.go
arjunrn: The way I see it the Obfuscator is the replacer here. For the case of random replacements we should not track what was replaced because that data is quite useless. For consistent/static replacement we should track replacements. And the replacements will always be the same. So if a key exists then it's value should always be used. It should not be overwritten. And it's not leaking abstractions because that's a desired property of the reporter(no dataloss).

tjungblu: I added that check as a panic, hope that works for you @arjunrn?

tjungblu: > The way I see it the Obfuscator is the replacer here. For the case of random replacements we should not track what was replaced because that data is quite useless. For consistent/static replacement we should track replacements. And the replacements will always be the same. So if a key exists then it's value should always be used. It should not be overwritten. And it's not leaking abstractions because that's a desired property of the reporter(no dataloss).

it's a matter of separation of concerns. The reporter should only report what I pass into it, not control what kind of values are permitted. 

>  For consistent/static replacement we should track replacements.

exactly, but who tracks that replacement? not the reporter I would say.

PR #13, opened 2021-08-10
Added a domain obfuscator which replaces all occurences of FQDNs

arjunrn:
The domain obfuscator changes all domain names such the root domain is rename but the subdomains are retained. Any root domains with TLDs are obfuscated along with the TLD.

Also added a configuration field to pass in TLDs.


pkg/schema/schema.json
```diff
@@ -75,6 +80,12 @@
                             }
                         ]
                     }
+                },
+                "topLevelDomains": {
+                    "type": "array",
```
tjungblu, line 85: you can supply:

```
     "type": "array",
      "items": {
          "type": "string"
      },
```

that seems to work with the `K8SOmission.Namespaces`, so should be supported by the generator.

pkg/cli/cli.go
```diff
@@ -25,6 +25,13 @@ func Run(configPath string, inputPath string, outputPath string) error {
 			obfuscators = append(obfuscators, obfuscator.NewKeywordsObfuscator(o.Replacement))
 		case schema.ObfuscateTypeMAC:
 			obfuscators = append(obfuscators, obfuscator.NewMacAddressObfuscator())
+		case schema.ObfuscateTypeDomain:
+			// The schema does not make the default value for the TLDs a string array instead an interface array.
```
tjungblu, line 29: yeah you can skip that when you supply the item types, left a comment below

pkg/obfuscator/domain.go
```diff
@@ -0,0 +1,109 @@
...
+	lock                 sync.Mutex
+	tlds                 map[string]struct{}
+	mainDomains          map[string]string
+	ipPattern            *regexp.Regexp
+	fileExtensionPattern *regexp.Regexp
+}
+
+func (d *domainObfuscator) FileName(s string) string {
+	// if the filename has an extension omit it from obfuscation
+	// this is still error-prone because the last part of the domain name could still be mistaken for an extension
```
tjungblu, line 23: isn't it a bit easier to take the `filepath.Ext` and just match on everything that comes before that? like: `s[: -len(filepath.Ext(s))]`. Would save a bit of regex matching

pkg/obfuscator/domain.go
```diff
@@ -0,0 +1,109 @@
...
+
+	parts := strings.Split(domain, ".")
+	var (
+		mainDomain   string
+		subDomain    string
+		hasExtension bool
+	)
+	// if the last part is present in known list of tlds then combine last and second-to-last parts to get
+	// the main domain otherwise the last part is the main domain
+	if _, ok := d.tlds[parts[len(parts)-1]]; ok {
```
tjungblu, line 61: if you found a domain that matches, can't we let the go url package parse it? that way you can just grab the hostname out of it.

pkg/obfuscator/domain.go
```diff
@@ -0,0 +1,109 @@
...
+	return output
+}
+
+func (d *domainObfuscator) Contents(s string) string {
+	return d.findDomains(s)
+}
+
+func (d *domainObfuscator) obfuscatedDomain(domain string) string {
+	d.lock.Lock()
+	defer d.lock.Unlock()
```
tjungblu, line 51: any reason for the locking yet? let's leave it out for now, the domainCount increment we can also do atomically without locking

pkg/schema/schema.json
```diff
@@ -75,6 +80,12 @@
...
+                    "title": "Top-Level-Domains",
+                    "description": "A list of TLDs which may be present in the input must-gather",
+                    "default": [".com", ".org", ".info", ".gov"]
```
tjungblu, line 88: is there a big TLD list that we can just checkin once?

tjungblu (changes requested):

pkg/schema/schema.json
arjunrn: But then the generated schema is not correct because the default values are of `[]interface{}` type.

pkg/cli/cli.go
arjunrn: Doesn't work, already tried.

pkg/schema/schema.json
tjungblu: the enums you mean? what a shame, I guess that's another patch for the other library for me then

pkg/obfuscator/domain.go
arjunrn: Wouldn't work, without the protocol the domain just parsed at the path.

pkg/schema/schema.json
arjunrn: There probably is, but there are many TLDs which are trademarked names and I didn't want to add them. Users can add whatever TLDs they use themselves.

tjungblu: well, we don't need to include `.fiat`, but can you please add the country codes and the remainders of the original ICANN set? https://en.wikipedia.org/wiki/Country_code_top-level_domain

pkg/schema/schema.json
tjungblu: https://github.com/atombender/go-jsonschema/pull/31

I guess we have to vendor my fork until we can have it in the "upstream" repo. I send a separate PR.

pkg/obfuscator/domain.go
arjunrn: Updated to use `filepath.Ext()`

pkg/obfuscator/domain.go
arjunrn: Removed.

pkg/cli/cli.go
```diff
@@ -25,6 +25,13 @@ func Run(configPath string, inputPath string, outputPath string) error {
...
+			var configTLDs []string
+			for _, tld := range config.Config.TopLevelDomains {
+				configTLDs = append(configTLDs, tld.(string))
+			}
+			obfuscators = append(obfuscators, obfuscator.NewDomainObfuscator(configTLDs))
```
tjungblu, line 34: if you want to rebase over #16 then you get that for free. The generator should generate the string types properly now. 

Let me know what you prefer, I can also fix it in #16 after we merge this PR

PR #15, opened 2021-08-11
Added obfuscator for IP Addresses

arjunrn:
The obfuscator uses regular expressions to detect strings which are like IPv4 and IPv6 addresses and then uses the ParseIP library to verify if the detected string is actually valid.


tjungblu:
thanks :)
/lgtm

PR #17, opened 2021-08-12
Refactored input logic for traversal.

arjunrn:
The input for the directory traversal is done through an interface with only the relative paths within the input directory. Previously the FileWalker was aware of the absolute paths in the input directory which made it more cumbersome to generate the name of the omitted files without the base input path.


pkg/input/input.go
```diff
@@ -0,0 +1,26 @@
+package input
+
+import (
+	"bufio"
+	"os"
+)
+
+type Directory interface {
```
tjungblu, line 8: all of those things already seem to exist in fs.go in the fs package, why do we need to reimplement that?

pkg/traversal/testfiles/test1/contents.yaml
```diff
@@ -1,6 +1,4 @@
 files:
-  "pods/manifests.yaml": "podkey:podcontents\n"
-  "nodes/test.yaml": "nodekey:nodecontents\n"
-directories:
-  "pods":
-  "nodes":
+  "pods/pod1/manifests.yaml": "podkey:podcontents\n"
+  "pods/pod2/manifests.yaml": "podkey:podcontents\n"
+  "nodes/test.yaml": "nodekey:nodecontents\n"
```
tjungblu, line 4: nit: newline

tjungblu (changes requested):

pkg/input/filesystem.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (f *fileSystemInput) Path() string {
+	return f.rootDir
+}
+
+func NewFSInput(inputDir string) (Inputter, error) {
+	fileInfo, err := os.Stat(inputDir)
+	if err != nil {
+		return nil, fmt.Errorf("cannot access reader directory %s: %w", inputDir, err)
```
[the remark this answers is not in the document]
arjunrn: The not-found error will be printed after cannot access.

pkg/input/filesystem.go
```diff
@@ -0,0 +1,116 @@
...
+func (f *fsFile) Path() string {
+	return f.path
+}
+
+func (f *fsFile) Name() string {
+	return f.fileInfo.Name()
+}
+
+func (f *fsFile) Mode() os.FileMode {
+	return f.fileInfo.Mode().Perm()
```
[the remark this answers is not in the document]
arjunrn: I've renamed it to `Permissions`

pkg/input/input.go
arjunrn: There's a `DirEntry` interface but it's not the same. I can rename the interface if you think it's confusing. I've refactored the structs to use some of interfaces but it's not a 1:1 match.

tjungblu: it's alright, the more we can reuse the better - in case we need other standard lib functionality that uses those interfaces. Thanks!

tjungblu:
let's get this in for now, we can refactor more later.

/lgtm

tjungblu (approved):

PR #18, opened 2021-08-12
Consistent replacements for obfuscated IP addresses

arjunrn:
Added a `GetReplacement()` method to the `ReplacementReporter` interface which returns a replacement if already present. This is leveraged by the IP obfuscator to find existing obfuscated values to consistently replace the same IP address with the same generated obfuscated value.

### DIFF
--- a/pkg/obfuscator/ip.go
+++ b/pkg/obfuscator/ip.go
@@ -1,15 +1,44 @@
 package obfuscator
 
 import (
+	"fmt"
 	"net"
 	"regexp"
 	"strings"
+
+	"github.com/openshift/must-gather-clean/pkg/schema"
 )
 
 const (
-	obfuscatedStaticIPv4 = "xxx.xxx.xxx.xxx"
-	obfuscatedStaticIPv6 = "xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx"
+	obfuscatedStaticIPv4         = "xxx.xxx.xxx.xxx"
+	obfuscatedStaticIPv6         = "xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx:xxxx"
+	consistentIPv4Template       = "x-ipv4-%06d-x"
+	consistentIPv6Template       = "xxxxxxxxxxxxx-ipv6-%06d-xxxxxxxxxxxxx"
+	maximumSupportedObfuscations = 999999
 )
+
+var (
+	ipv6Pattern = regexp.MustCompile(ipv6re)
+	ipv4Pattern = regexp.MustCompile(ipv4re)
+)
+
+type ipGenerator struct {
+	template   string
+	obfuscated string
+	count      int
+}
+
+func (g *ipGenerator) generateConsistent() string {
+	g.count++
+	if g.count > maximumSupportedObfuscations {
+		panic("maximum number of obfuscated ips exceeded")
+	}
+	return fmt.Sprintf(g.template, g.count)
+}
+
+func (g *ipGenerator) static() string {
+	return g.obfuscated
+}
 
 var (
 	ipv4re = `(([0-9]{1,3})\.){3}([0-9]{1,3})`
@@ -19,8 +48,8 @@ var (
 
 type ipObfuscator struct {
 	ReplacementReporter
-	ipv4pattern *regexp.Regexp
-	ipv6pattern *regexp.Regexp
+	replacements    map[*regexp.Regexp]ipGenerator
+	replacementType schema.ObfuscateReplacementType
 }
 
 func (o *ipObfuscator) FileName(s string) string {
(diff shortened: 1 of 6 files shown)


pkg/obfuscator/ip.go
```diff
@@ -36,26 +44,51 @@ func (o *ipObfuscator) replace(s string) string {
...
-			output = strings.ReplaceAll(output, m, obfuscatedStaticIPv4)
-			o.ReportReplacement(m, obfuscatedStaticIPv4)
+			var replacement string
+			switch o.replacementType {
+			case schema.ObfuscateReplacementTypeStatic:
+				replacement = obfuscatedStaticIPv4
+			case schema.ObfuscateReplacementTypeConsistent:
+				if replacement = o.HasReplacement(m); replacement == "" {
+					o.ipv4count++
+					replacement = fmt.Sprintf(consistentIPv4Template, o.ipv4count)
```
tjungblu, line 54: we don't need to solve this in this PR, but we need to make it look like a proper IP address - ideally with some random octets instead of a counter.

pkg/obfuscator/ip.go
```diff
@@ -36,26 +44,51 @@ func (o *ipObfuscator) replace(s string) string {
...
+				}
+			default:
+				replacement = m
```
tjungblu, line 57: should this actually happen?

pkg/obfuscator/ip.go
```diff
@@ -36,26 +44,51 @@ func (o *ipObfuscator) replace(s string) string {
...
+			case schema.ObfuscateReplacementTypeConsistent:
+				if replacement = o.HasReplacement(m); replacement == "" {
+					o.ipv6count++
+					replacement = fmt.Sprintf(consistentIPv6Template, o.ipv6count)
+				}
+			default:
+				replacement = m
+			}
+			output = strings.ReplaceAll(output, m, replacement)
+			o.ReportReplacement(m, replacement)
```
tjungblu, line 80: can you refactor those two V4+V6 code blocks into one method? I think you only differ in the use of the regex, the template and the counter

tjungblu: + same question for the default switch, probably should be an error

pkg/obfuscator/ip.go
```diff
@@ -36,26 +44,51 @@ func (o *ipObfuscator) replace(s string) string {
...
 		}
 	}
 
 	return output
 }
 
-func NewIPObfuscator() Obfuscator {
+func NewIPObfuscator(replacementType schema.ObfuscateReplacementType) Obfuscator {
 	return &ipObfuscator{
```
tjungblu, line 88: please add some validation that we only take in replacement types that this obfuscator actually supports. Then your default case can just panic or return errors.

pkg/obfuscator/ip_test.go
```diff
@@ -59,10 +61,81 @@ func TestIPObfuscator(t *testing.T) {
...
+		name   string
+		input  []string
+		output []string
+		report map[string]string
+	}{
+		{
+			name:   "valid ipv4 address",
+			input:  []string{"received request from 192.168.1.10"},
+			output: []string{"received request from x-ipv4-000001-x"},
+			report: map[string]string{"192.168.1.10": "x-ipv4-000001-x"},
```
tjungblu, line 83: what's your overflow behaviour for your template pattern "x-ipv4-000001-x"?  maybe add a test that starts at `x-ipv4-999999-x` and check the result? 

probably warrants for a proper error message before the "Sprintf"

pkg/obfuscator/replacement_reporter.go
```diff
@@ -12,6 +12,9 @@ type ReplacementReporter interface {
 	// ReportReplacement will add a replacement along with its original string to the report.
 	// If there is an existing value that does not match the given replacement, it will panic as this very likely denotes a bug.
 	ReportReplacement(original string, replacement string)
+
+	// HasReplacement returns the previously used replacement if already set.
+	HasReplacement(original string) string
```
tjungblu, line 17: so it's a `GetReplacement`? please also document the behaviour if it doesn't exist and add tests.

tjungblu (changes requested):

pkg/obfuscator/ip.go
arjunrn: Why? What's the benefit of doing that?

pkg/obfuscator/ip.go
arjunrn: It shouldn't but this guard condition prevents the obfuscated output from being blank.

pkg/obfuscator/ip.go
tjungblu: actually you're right, I thought it was in the spec like that.

pkg/obfuscator/ip_test.go
arjunrn: Since I don't expect there will be more than a million IP addresses to obfuscate I've added a panic when the counter would overflow. We can also not have the panic because if the number exceeds the formatting directive then the formatting directive is ignored.

pkg/obfuscator/ip.go
arjunrn: Done

tjungblu (approved):
good to go, thanks!

/lgtm

PR #19, opened 2021-08-16
File omission based on file glob

arjunrn:
The file omitter can use the same pattern to also match against full paths instead of just the filename.

This PR requires #17 so that the relative paths from within the input directory are available within the `FileWalker`.

Also removes the `/` from the input reader so that filepath globs without the leading slash can match.


tjungblu:
easy :) Thanks for adding it.

/lgtm

arjunrn:
/hold
Till #17 is merged.
/lgtm cancel

PR #20, opened 2021-08-18
Refactored the omit configuration schema

arjunrn:
Currently the omit coniguration property can have multiple object types as a value. As a result the type of the `omit` property is `[]interface{}` which makes validation and initialization of the omitters more difficult. This change refactor the omit `type` to use a single type with multiple configuration fields which can be validated in the application.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -44,9 +44,12 @@ func Run(configPath string, inputPath string, outputPath string) error {
 
 	var omitters []omitter.Omitter
 	for _, o := range config.Config.Omit {
-		switch om := o.(type) {
-		case schema.FileOmission:
-			omitters = append(omitters, omitter.NewFilenamePatternOmitter(om.Pattern))
+		switch o.Type {
+		case schema.OmitTypeFile:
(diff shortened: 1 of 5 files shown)


pkg/cli/cli.go
```diff
@@ -44,9 +44,12 @@ func Run(configPath string, inputPath string, outputPath string) error {
 
 	var omitters []omitter.Omitter
 	for _, o := range config.Config.Omit {
-		switch om := o.(type) {
-		case schema.FileOmission:
-			omitters = append(omitters, omitter.NewFilenamePatternOmitter(om.Pattern))
+		switch o.Type {
+		case schema.OmitTypeFile:
+			if *o.Pattern == "" {
```
tjungblu, line 49: nit: I thought we wanted to add the validations into the NewXX funcs

tjungblu (approved):
rest looks good

pkg/cli/cli.go
arjunrn: Fixed in [follow-up](https://github.com/openshift/must-gather-clean/pull/21)

PR #21, opened 2021-08-18
Validate the input for file omitter in New function

arjunrn:
Validate the data from the configuration in the `NewFilePatternOmitter()` function instead of the `cli` package.


tjungblu:
thx!!
/lgtm

PR #22, opened 2021-08-18
Added reporting of replacements and omissions

arjunrn:
All replacements done by the obfuscators and files which are omitted are included in a report which is written to the root of the output directory.


tjungblu (approved):
/lgtm

PR #24, opened 2021-08-19
Kubernetes Resource omitter

arjunrn:
The resource omitter reads the contents of a json or yaml file and tries to decode them as a kubernetes resource or a resource list. For a resource list if the input criteria of kind, api version and namespaces match then the omitter indicates this. If for a resource list one of the item matches the criteria then the entire file is omitted.

/assign @tjungblu

### DIFF
--- a/pkg/omitter/kubernetesresource.go
+++ b/pkg/omitter/kubernetesresource.go
@@ -1,0 +1,114 @@
+package omitter
+
+import (
+	"encoding/json"
+	"errors"
+	"io/ioutil"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+)
+
+type kubeMetadata struct {
+	Namespace string `yaml:"namespace" json:"namespace"`
+}
+
+type kubeResource struct {
+	ApiVersion string       `yaml:"apiVersion" json:"apiVersion"`
+	Kind       string       `yaml:"kind" json:"kind"`
+	Metadata   kubeMetadata `yaml:"metadata" json:"metadata"`
+}
+
+type kubeResourceList struct {
+	Items []kubeResource `yaml:"items" json:"items"`
+}
+
+type resourceUnmarshaller func(in []byte, out interface{}) (err error)
+
+type kubernetesResourceOmitter struct {
+	apiVersion   string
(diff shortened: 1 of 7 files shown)


pkg/omitter/kubernetesresource.go
```diff
@@ -0,0 +1,114 @@
...
+		resourceList = kubeResourceList{Items: []kubeResource{resource}}
+	}
+
+	if len(resourceList.Items) == 0 {
+		return false, nil
+	}
+
+	var found bool
+	// loop over the resources and if one of them matches the criteria then set the `found` flag.
+	for _, r := range resourceList.Items {
```
tjungblu, line 77: whole thing /lgtm, what do you think of pulling this whole k8s stuff above a level higher into the traversal? 
Ideally we should only obfuscate the values of the yaml and not a line-based version of it, at least that was the superpower over sos-cleaner. We can then also build specific obfuscators and omitters based on that schema - think of a configmap filter that looks for certificates in values which is not that easy line-by-line.

tjungblu (approved):

pkg/omitter/kubernetesresource.go
arjunrn: Omitters read the entire file because they need the whole file to determine if something is valid kubernetes resource. We can have more complex filter of course but that's not omission, it's more obfuscation.

tjungblu: exactly, nothing to fix right here. I'm just thinking about pulling the parsing logic a level up, so we can offer a schema/type supported way of `Obfuscator.Contents`. Omission can then leverage the same interface.

basically an addition to the line-by-line processor here: https://github.com/openshift/must-gather-clean/blob/main/pkg/traversal/traversal.go#L108-L128

arjunrn: I get it now. So there could be obfuscators which could work on parts of the resource schema this way. It's possible. But I would like to see more user-stories before we commit to something.

tjungblu: sounds good, I'll write some tomorrow

PR #26, opened 2021-08-19
add overwrite flag for the output folder

tjungblu:


pkg/output/output.go
```diff
@@ -48,29 +48,46 @@ func (f *fsWriter) Writer(parentPath, name string, permissions os.FileMode) (Clo
...
-// NewFSWriter returns an Outputter which writes the files and directories to a specified location.
-func NewFSWriter(path string) (Outputter, error) {
+func EnsureOutputPath(path string, deleteIfExists bool) error {
 	info, err := os.Stat(path)
 	if err != nil {
 		if !os.IsNotExist(err) {
-			return nil, err
+			return err
 		} else {
 			err := os.Mkdir(path, 0700)
```
arjunrn, line 57: You can return at this point and get rid of the else branch.

tjungblu: done!

pkg/output/output.go
```diff
@@ -48,29 +48,42 @@ func (f *fsWriter) Writer(parentPath, name string, permissions os.FileMode) (Clo
...
+func EnsureOutputPath(path string, deleteIfExists bool) error {
 	info, err := os.Stat(path)
 	if err != nil {
-		if !os.IsNotExist(err) {
-			return nil, err
+		if os.IsNotExist(err) {
+			return os.Mkdir(path, 0700)
+		}
+		return err
+	} else {
```
arjunrn, line 58: You don't need to have the `else` branch anymore. Or rather whatever is in the else branch can be after the `if`. So the steps would be:
1. Stat the input path.
2. If error check if not exists and create if it does not. Either way return.
3. From stat info check if input is directory. if not return error. We don't want to accidentally delete something which is not an output directory.
4. check if directory is empty.
4.1 if not empty and delete flag is set delete
4.2 else return error
5. create directory
6. return

tjungblu: done

pkg/output/output.go
```diff
@@ -55,31 +55,33 @@ func EnsureOutputPath(path string, deleteIfExists bool) error {
...
-				return fmt.Errorf("output destination must be a directory: '%s'", path)
-			}
-			entries, err := os.ReadDir(path)
-			if err != nil {
-				return fmt.Errorf("failed to get contents of output directory '%s': %w", path, err)
-			}
-			if len(entries) != 0 {
-				return fmt.Errorf("output directory %s is not empty", path)
-			}
+			return fmt.Errorf("output directory %s is not empty", path)
```
arjunrn, line 76: This should be after the `if deleteIfExists{}`

tjungblu: it is after the `deleteIfExists`. or do you mean to just leave the else statement out?

arjunrn: No this fine. I was reviewing in between updates.

PR #29, opened 2021-08-19
Allow dashes in IPv4 address pattern

arjunrn:
Some file names may contain IPv4 addresses separated by dashes. The obfuscator should take that into account when detecting valid IPv4 addresses and use any existing replacements into account.


tjungblu (approved):
/lgtm

PR #30, opened 2021-08-20
Only obfuscated user supplied domains

arjunrn:
This change adds a new field in the obfuscation type of `Domain` which is used to obfuscated only domains and subdomains of the input.


pkg/schema/schema.json
```diff
@@ -109,6 +100,14 @@
                     ],
                     "description": "type defines the kind of detection you want to use. For example IP will find IP addresses, whereas Keywords will find predefined keywords."
                 },
+                "domains": {
+                    "description": "The list of domains and their subdomains which should be obfuscated in the output.",
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "minItems": 1
```
tjungblu, line 108: hmm, it didn't generate that check- I'll debug it.

pkg/obfuscator/domain.go
```diff
@@ -2,102 +2,77 @@ package obfuscator
...
 import (
 	"fmt"
-	"path/filepath"
 	"regexp"
 	"strings"
 )
 
+const (
+	domainPattern      = `([a-zA-Z0-9\.]*\.)?(%s)`
+	obfuscatedTemplate = "domain%07d"
```
tjungblu, line 11: aside that we now replace it with this pattern, can't we solve the whole thing with the keyword obfuscator?

tjungblu: ah no, we're doing additional logic for detecting the sub domains. nvm!

pkg/obfuscator/domain.go
```diff
@@ -2,102 +2,77 @@ package obfuscator
...
+	d.domainMapping[domain] = replacement
+	return replacement
 }
 
-func NewDomainObfuscator(tlds []string) Obfuscator {
-	tldSet := map[string]struct{}{}
-	for _, tld := range tlds {
-		tldSet[strings.TrimLeft(tld, ".")] = struct{}{}
+func NewDomainObfuscator(domains []string) (Obfuscator, error) {
+	patterns := make([]*regexp.Regexp, len(domains))
```
tjungblu, line 64: now that we have the static and consistent types, can you also add that behaviour in here?

pkg/obfuscator/domain_test.go
```diff
@@ -3,106 +3,61 @@ package obfuscator
...
-			name: "subdomain",
-			tlds: []string{".com", ".info", ".org"},
-			input: []string{
-				"received request on abc.test.com",
-				"received request on def.test.com",
-			},
-			output: []string{
-				"received request on abc.obfuscated0001.ext",
-				"received request on def.obfuscated0001.ext",
+				"received request on domain0000001",
```
tjungblu, line 27: I'm actually wondering whether we can keep the TLD so it at least looks like a domain.

pkg/obfuscator/domain_test.go
```diff
@@ -3,106 +3,61 @@ package obfuscator
...
 			output: []string{
-				"received request for https://service.obfuscated0001.ext from 192.168.1.20",
+				"received request on abc.domain0000001",
+				"received request on def.domain0000002",
+				"received request on ghi.abc.domain0000001",
+				"received request on pqr.ghi.abc.domain0000001",
 			},
 			report: map[string]string{
-				"service.example.com": "service.obfuscated0001.ext",
+				"abc.test.com":         "abc.domain0000001",
```
tjungblu, line 51: can you also add a test around those subdomains? 
so what if I wanted to have `abc.test.com` obfuscated, instead of just the `test.com`.

pkg/obfuscator/domain_test.go
```diff
@@ -114,34 +69,32 @@ func TestDomainObfuscatorContents(t *testing.T) {
...
+			report:  map[string]string{},
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
-			o := NewDomainObfuscator(nil)
+			o, err := NewDomainObfuscator(tc.domains)
+			require.NoError(t, err)
 			output := o.FileName(tc.input)
 			assert.Equal(t, tc.output, output)
 			assert.Equal(t, tc.report, o.ReportingResult())
```
tjungblu, line 100: please also add a test for when that regex doesn't compile it should return an error

tjungblu (changes requested):
looks great already, I'll checkout that minItems bug in the generator

pkg/schema/schema.json
tjungblu: ah it's in the wrong level, you need to pull it up one so it's side-by-side with the items

pkg/obfuscator/domain.go
arjunrn: In a separate PR. Right now this is only consistent replacement.

pkg/obfuscator/domain_test.go
arjunrn: The `soscleaner` doesn't. I don't see how it would be help anyway.

tjungblu: with readability maybe? keep in mind that you also need to be on the receiving end of that must-gather at some point and that would make it look more naturally like a domain. In VSCode at least it gives me a domain/link highlighting.

pkg/schema/schema.json
arjunrn: I think the generated validation is still broken because it expects the field to always be of length 1, even when not specified.

tjungblu: not sure it's properly defined in the RFC http://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.4.2

I'd leave it out then, if we desire the domains to be optional.

pkg/obfuscator/domain_test.go
arjunrn: I believe the highlighting is because it contains a protocol prefix like `http://` or `https://` which will still be there.

tjungblu: that's the underlining and making it clickable, but it also highlights things like quay.io properly without protocols. Anyway, why remove the ext if it was there before?

tjungblu: talking about quay.io, shall we also keep an inclusion list of domains that are not private like openshift.io/com and quay?

arjunrn: I removed it because it makes the logic more complex without substantial benefits. Because users might want to obfuscate domains which don't have extensions I don't think they would want it indicated if it has an extension or not.

arjunrn: Since we have changed from a block-list to allow-list approach if the users don't want `openshift.io` they simply leave it out from the domains list.

pkg/obfuscator/domain.go
```diff
@@ -2,102 +2,77 @@ package obfuscator
...
 	return d.findDomains(s)
 }
 
-func (d *domainObfuscator) findDomains(input string) string {
-	domains := d.hostPattern.FindAllString(input, -1)
+func (d *domainObfuscator) Contents(s string) string {
+	return d.findDomains(s)
+}
 
+func (d *domainObfuscator) findDomains(input string) string {
```
tjungblu, line 29: does the naming still make sense? it replaces the domains it found

pkg/obfuscator/domain_test.go
tjungblu: :+1: we can add it back if that's confusing for real users then.

pkg/obfuscator/domain.go
arjunrn: Changed to `replaceDomains`.

pkg/obfuscator/domain_test.go
arjunrn: Done.

pkg/obfuscator/domain_test.go
arjunrn: Added a test

tjungblu:
cool, /lgtm thanks!

PR #32, opened 2021-08-20
introduce proper logging framework, replace panics and printfs

tjungblu:


pkg/obfuscator/ip_test.go
```diff
@@ -205,15 +205,3 @@ func TestIPObfuscatorConsistent(t *testing.T) {
 		})
 	}
 }
-
-func TestPanicMaximumReplacements(t *testing.T) {
```
tjungblu, line 209: that's sadly not testable due to the exit, but I think it's the overall better behaviour than the panic

pkg/traversal/traversal.go
```diff
@@ -84,13 +85,19 @@ func (w *FileWalker) processDir(inputDir input.Directory, outputDirName string)
 			}
 			// If the file should be omitted then stop processing.
 			if omit {
+				if klog.V(2).Enabled() {
+					klog.Infof("omitting '%s'", e.Path())
```
arjunrn, line 89: Can't this just be `klog.V(2).Infof()`?

pkg/traversal/traversal.go
```diff
@@ -129,6 +136,10 @@ func (w *FileWalker) processDir(inputDir input.Directory, outputDirName string)
 			if err != nil {
 				return err
 			}
+
+			if klog.V(2).Enabled() {
+				klog.Infof("done processing '%s'", e.Path())
```
arjunrn, line 141: And here.

pkg/traversal/traversal.go
tjungblu: ah you're right, I was looking at the docs the wrong way. I thought the latter wouldn't skip the locking. Fixing

PR #33, opened 2021-08-20
fix in reporting, mkdirs, docs + example

tjungblu:


pkg/cli/cli.go
```diff
@@ -97,15 +100,16 @@ func Run(configPath string, inputPath string, outputPath string, deleteOutputFol
 	}
 
 	report := walker.GenerateReport()
-	rPath := filepath.Join(outputPath, reportFileName)
-	reportFile, err := os.Create(rPath)
+	// we store the report in the working dir of the CLI, it should never be created in the output folder
```
arjunrn, line 103: This comment describes the change being made is not actually useful when read by itself.

arjunrn: Also if you want to store it in another location then use a flag which the user can set.

tjungblu: it was more intended to be a warning for others changing. But it totally makes sense to add a flag for that, will do later.

tjungblu: done

PR #34, opened 2021-08-20
Concurrent processing of input directory

arjunrn:
This change modifies the FileWalker to use workers which can process files concurrently. Files are passed to the workers through channels and an error channel returns any errors which might have been encountered when processing a file. When the error channel returns an error the FileWalker immediately exits with the received error..

### DIFF
--- a/pkg/obfuscator/ip.go
+++ b/pkg/obfuscator/ip.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/openshift/must-gather-clean/pkg/schema"
 	"k8s.io/klog/v2"
@@ -32,27 +33,60 @@ var (
 )
 
 type ipGenerator struct {
-	template   string
-	obfuscated string
-	count      int
+	ipReplacements map[string]string
+	template       string
+	obfuscated     string
+	count          int
+	lock           sync.RWMutex
 }
 
-func (g *ipGenerator) generateConsistent() string {
+func (g *ipGenerator) consistent(ip string) string {
+	g.lock.Lock()
+	defer g.lock.Unlock()
+
+	if r, ok := g.ipReplacements[ip]; ok {
+		return r
+	}
+
 	g.count++
 	if g.count > maximumSupportedObfuscations {
 		klog.Exitf("maximum number of ip obfuscations exceeded: %d", maximumSupportedObfuscations)
 	}
-	return fmt.Sprintf(g.template, g.count)
+	r := fmt.Sprintf(g.template, g.count)
+	g.ipReplacements[ip] = r
+	return r
 }
 
-func (g *ipGenerator) static() string {
+func (g *ipGenerator) static(ip string) string {
+	g.lock.Lock()
+	defer g.lock.Unlock()
+	g.ipReplacements[ip] = g.obfuscated
 	return g.obfuscated
 }
 
+func (g *ipGenerator) replacements() map[string]string {
+	g.lock.RLock()
+	defer g.lock.RUnlock()
+	rCopy := make(map[string]string)
+	for k, v := range g.ipReplacements {
+		rCopy[k] = v
+	}
+	return rCopy
+}
+
 type ipObfuscator struct {
-	ReplacementReporter
 	replacements    map[*regexp.Regexp]*ipGenerator
 	replacementType schema.ObfuscateReplacementType
+}
+
+func (o *ipObfuscator) ReportingResult() map[string]string {
+	result := make(map[string]string)
+	for _, replacers := range o.replacements {
+		for k, v := range replacers.replacements() {
+			result[k] = v
+		}
+	}
+	return result
 }
 
 func (o *ipObfuscator) FileName(s string) string {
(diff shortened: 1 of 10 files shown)


tjungblu:
> The errors are now simply printed out as warnings.

keep in mind that we should exit the tool with non-zero code in case we had some errors.

pkg/obfuscator/ip.go
```diff
@@ -35,9 +36,12 @@ type ipGenerator struct {
 	template   string
 	obfuscated string
 	count      int
+	lock       sync.Mutex
 }
 
 func (g *ipGenerator) generateConsistent() string {
+	g.lock.Lock()
+	defer g.lock.Unlock()
```
tjungblu, line 44: I can't comment down below because that's out of the diff, but I do believe there is a race with the reporter:

```
				case schema.ObfuscateReplacementTypeConsistent:
					if replacement = o.GetReplacement(m); replacement == "" {
						replacement = gen.generateConsistent()
					}
				}
				output = strings.ReplaceAll(output, m, replacement)
				o.ReportReplacement(m, replacement)
```

you can run into a scenario where you find there was no replacement in two threads, generate two different replacements and then report both later which will exit with an error while reporting the second one.

Go should pick this up with a race detector actually :octopus:

arjunrn: Indeed, I've made some changes which should fix the issue.

pkg/obfuscator/ip.go
```diff
@@ -32,29 +33,50 @@ var (
...
 	}
-	return fmt.Sprintf(g.template, g.count)
+	r := fmt.Sprintf(g.template, g.count)
+	g.replacements[ip] = r
+	return r
 }
 
-func (g *ipGenerator) static() string {
+func (g *ipGenerator) static(ip string) string {
+	g.replacements[ip] = g.obfuscated
```
tjungblu, line 61: I think you also need to put this under the lock - what if the hashmap underneath rehashes the entries?

pkg/obfuscator/ip.go
```diff
@@ -32,29 +33,50 @@ var (
...
 
 type ipObfuscator struct {
-	ReplacementReporter
 	replacements    map[*regexp.Regexp]*ipGenerator
 	replacementType schema.ObfuscateReplacementType
 }
 
+func (o *ipObfuscator) ReportingResult() map[string]string {
+	result := make(map[string]string)
+	for _, replacers := range o.replacements {
```
tjungblu, line 72: please also put that under the lock, you need the memory barrier otherwise you can read some funny cached mapping

pkg/obfuscator/obfuscator.go
```diff
@@ -2,9 +2,10 @@ package obfuscator
 
 // Obfuscator is the interface which all obfuscators should implement
 type Obfuscator interface {
-	ReplacementReporter
 	// FileName takes a filename as input and return the obfuscated name
 	FileName(string) string
 	// Contents takes string as input and return the obfuscated string
 	Contents(string) string
+	// ReportingResult returns a map of words and their replacements
+	ReportingResult() map[string]string
```
tjungblu, line 10: interesting choice, why not just refactor the `ReplacementReporter` to have an atomic report+get?

tjungblu: or have an atomic replacement reporter and kick the simple one out of the project

pkg/obfuscator/ip.go
arjunrn: I didn't see an issue with this because this is called after the processing is actually done. But I see your point.

pkg/obfuscator/obfuscator.go
arjunrn: It's not that easy. The reporter would need to have a generator which generates a replacement when there is no existing replacement. Because the new replacement should be generated atomically as well. So I move the entire logic into the `ipGenerator`. Also IMO it doesn't make sense for an obfuscator to expose methods like `ReportReplacement`.

pkg/obfuscator/ip.go
arjunrn: Done

pkg/obfuscator/obfuscator.go
tjungblu: what if you pass a function for the generation? eg like https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent(K,java.util.function.Function)

arjunrn: That's possible. But I don't like that then the `GetReplacement` will return a replacement in the generating reporter and in the `SimpleReporter` it won't. If we want to refactor this we should do it separately.

tjungblu: okay cool, I will send you a PR then with the refactoring later today. let's get this merged :)

pkg/obfuscator/ip.go
arjunrn: Updated

tjungblu (approved):
/lgtm

PR #35, opened 2021-08-23
Exclude common non-sensitive IPs from obfuscation

arjunrn:
Certain IPs like 127.0.0.1 are not sensitive information and should not be obfuscated. They may be necessary for accurate debugging.


tjungblu (approved):
/lgtm

PR #37, opened 2021-08-24
Refactor of replacment reporting

arjunrn:
This change refactors the replacement reporting interface to define new methods which are used to atomically initialize the replacements tracker and also generate replacements for missing keys.


pkg/obfuscator/ip.go
```diff
@@ -33,62 +34,34 @@ var (
...
 type ipGenerator struct {
-	ipReplacements map[string]string
-	template       string
-	obfuscated     string
-	count          int
-	lock           sync.RWMutex
+	template   string
+	obfuscated string
+	count      int
+	lock       sync.RWMutex
```
tjungblu, line 40: since the critical sections are now in the `ReplacementReporter`, I think you can completely remove this one. There can only be one thread ever incrementing the counter.

pkg/obfuscator/tracker.go
```diff
@@ -0,0 +1,84 @@
...
+	AddReplacement(original string, replacement string)
+
+	// GetReplacement returns the previously used replacement if already set. If the replacement is not present then it
+	// uses the GenerateReplacement function to generate a replacement. If the generator is nil then it returns an empty
+	// string. The original parameter must be used for lookup and the key parameter to generate the replacement.
+	GetReplacement(original string, key string, generator GenerateReplacement) string
+
+	// Initialize initializes the tracker with some existing replacements. It should be called only once and before
+	// the first use of GetReplacement or AddReplacement
+	Initialize(replacements map[string]string)
```
tjungblu, line 27: super nit pick, add that to the top of the type, so people know that this needs to be initialized first.

pkg/obfuscator/tracker.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"sync"
+
+	"k8s.io/klog/v2"
+)
+
+type GenerateReplacement func(string) string
+
+// ReplacementTracker is used to track and generate replacements used by obfuscators
+type ReplacementTracker interface {
```
tjungblu, line 12: interface looks great, good job

pkg/obfuscator/tracker.go
```diff
@@ -0,0 +1,84 @@
...
+	Report() map[string]string
+
+	// AddReplacement will add a replacement along with its original string to the report.
+	// If there is an existing value that does not match the given replacement, it will exit with a non-zero status.
+	AddReplacement(original string, replacement string)
+
+	// GetReplacement returns the previously used replacement if already set. If the replacement is not present then it
+	// uses the GenerateReplacement function to generate a replacement. If the generator is nil then it returns an empty
+	// string. The original parameter must be used for lookup and the key parameter to generate the replacement.
+	GetReplacement(original string, key string, generator GenerateReplacement) string
```
tjungblu, line 23: do you mind if we stick to the java nomenclature of PutIfAbsent?

pkg/obfuscator/ip.go
arjunrn: Yes, in this context it's redundant but I would like to keep it there so that in future anyone making changes here knows that it should be thread-safe.

pkg/obfuscator/tracker.go
arjunrn: The Java name is such because there a a ton of functions like `Get`, `GetIfAbsent`, `GetIfPresent` and so on. But in this case there is only one function so `GetReplacement` is appropriate IMO.

pkg/obfuscator/tracker.go
arjunrn: Done

pkg/obfuscator/tracker.go
tjungblu: maybe GenerateIfAbsent? It certainly does more than just getting something

pkg/obfuscator/ip.go
tjungblu: let's rather leave a comment that this would be covered by the ReplacementReporter's locking, I foresee more confusion and deadlock potential with two locks involved

pkg/obfuscator/tracker.go
arjunrn: But it also does a simple `Get` if the generator is `nil`.

tjungblu: shall we split it into two funcs then, one just doing a get and one with a generator `GenerateIfAbsent` maybe?

arjunrn: Why?

pkg/obfuscator/ip.go
arjunrn: Done

pkg/obfuscator/tracker.go
tjungblu: Separation of concerns I'd say. One is just doing a Get that returns empty when there is nothing, otherwise the result. The other is atomically replacing a key if there is no result and returns that. It's two different things that we shouldn't really mix IMHO.

arjunrn: Renamed.

tjungblu: thanks, apologies for bikeshedding

PR #43, opened 2021-08-31
refactor io system and reporting

tjungblu:
I simplified the IO logic a bit and split the traversal and actual worker out to separate the concerns a bit better. I also added an easier reporting facility for Ramesh to pick up.

cc @pweil- if you have some time


pkg/omitter/fileomitter.go
```diff
@@ -9,14 +9,8 @@ type filePatternOmitter struct {
...
-func (f *filePatternOmitter) Omit(name, path string) (bool, error) {
-	match, err := filepath.Match(f.filePattern, path)
-	// if there is a match or an error return now
-	if match || err != nil {
-		return match, err
-	}
-	// else check for match with file name
-	return filepath.Match(f.filePattern, name)
+func (f *filePatternOmitter) Omit(path string) (bool, error) {
+	return filepath.Match(f.filePattern, path)
```
arjunrn, line 13: This won't match patterns like `*.log`. That was part of the spec right?

pkg/traversal/traversal.go
```diff
@@ -1,59 +1,33 @@
...
+	queue := make(chan WorkerInput, w.workerCount)
+	w.workers = make([]QueueProcessor, w.workerCount)
 	for i := 0; i < w.workerCount; i++ {
-		wk := newWorker(i+1, w.obfuscators, w.fileOmitters, w.k8sOmitters, queue, w.writer, errorCh)
-		w.workers[i] = wk
+		w.workers[i] = w.workerFactory(i + 1)
 		wg.Add(1)
-		go func() {
-			wk.run()
+		go func(i int, queue chan WorkerInput, errorCh chan error) {
```
arjunrn, line 27: Does `WorkerInput` have to be a public type? Also what's the reason for passing them in here in `ProcessQueue()` instead of the new function?

pkg/traversal/traversal.go
```diff
@@ -71,7 +45,32 @@ func (w *FileWalker) Traverse() {
...
+	err := filepath.WalkDir(w.inputPath, func(path string, dirEntry fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if !dirEntry.IsDir() {
+			// the rest of the logic expects the path to be relative to the input dir root, if it fails we assume it is already relative
+			relPath, err := filepath.Rel(w.inputPath, path)
+			if err != nil {
+				queue <- WorkerInput{
```
arjunrn, line 57: When there is an error shouldn't this be written into the error channel?

arjunrn: When and how would it fail?

pkg/traversal/traversal_test.go
```diff
@@ -1,122 +1,49 @@
...
-		inputDir string
+		name           string
+		inputDir       string
+		expectedResult []string
 	}{
 		{
 			name:     "basic",
-			inputDir: "testfiles/test1",
+			inputDir: "testfiles/test1/mg",
+			expectedResult: []string{
```
arjunrn, line 28: It appears the tests are not using the reports file anymore. They should be removed.

pkg/traversal/worker.go
```diff
@@ -25,69 +26,54 @@ func (f *fileProcessingError) Cause() error {
 	return f.cause
 }
 
-type workerFile struct {
-	f input.File
+type WorkerInput struct {
```
arjunrn, line 29: Since `WorkerInput` holds just one string we can turn use strings or a type aliased to string.

pkg/traversal/worker.go
```diff
@@ -159,42 +145,77 @@ func (w *worker) shouldOmitK8sResource(resource *kube.ResourceList) (bool, error
...
 		}
-		_, err = writer.WriteString(fmt.Sprintf("%s\n", contents))
+
+		_, err = fmt.Fprintln(writer, contents)
 		if err != nil {
 			return err
 		}
 	}
+
+	// we deliberately do not defer the closes, as this will be processed asynchronous and
```
arjunrn, line 182: This comment does not make clear to anyone reading the code. Because returning an error makes the error reader worker call `klog.Exitf()` and hence all file handles will immediately be cleaned up.

pkg/omitter/fileomitter.go
tjungblu: schema says: 

> A file glob pattern on file paths relative to the must-gather root. The pattern should be as described in https://pkg.go.dev/path/filepath#Match

https://github.com/openshift/must-gather-clean/blob/main/pkg/schema/schema.json#L208

pkg/traversal/traversal_test.go
tjungblu: which reports? they are now in a dedicated struct

pkg/traversal/worker.go
tjungblu: yeah I thought about this as well, but if we add more context information later on it's not as big of a refactoring. Whatever you prefer.

pkg/omitter/fileomitter.go
arjunrn: But the example says https://github.com/openshift/must-gather-clean/blob/e03633880b9c93bfbffbd63d3d647fa0369881f8/pkg/schema/schema.json#L95

pkg/traversal/traversal_test.go
arjunrn: This one https://github.com/openshift/must-gather-clean/blob/main/pkg/traversal/testfiles/test1/report.yaml

pkg/omitter/fileomitter.go
tjungblu: it does, but that matches log files in the must gather root folder, which there aren't any as of today. But agreed, that example is confusing, will rewrite it.

pkg/traversal/reporter.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// ReportOmission saves the given path as omitted to later report as such.
+	ReportOmission(path string)
+
+	// ReportObfuscators will call the Report method on all obfuscators and collect the individual obfuscation results for final reporting.
+	ReportObfuscators(obfuscators []obfuscator.Obfuscator)
+}
+
+type SimpleReporter struct {
+	lock         *sync.RWMutex
```
pweil-, line 31: see how this does but one thing to keep in mind with concurrent code is that you can often avoid locks altogether by using immutable objects.  Ie. an obfuscator returns an immutable report result which you then add to your private aggregate.  Same for omissions.

If the overhead of locking is noticeable that might be something to give a go.

pkg/omitter/fileomitter.go
arjunrn: This was in the design doc:
```
   # this omits all log files
   - type: file
     pattern: "*.log"
```

pkg/traversal/worker.go
arjunrn: Let's use a type aliased to string.

pkg/traversal/reporter.go
pweil-: it's good to leave a comment on what the lock is intended to guard.  It looks like this case it's that the entire type, like an intrinsic lock in java

arjunrn:
JFYI the abstractions for the input and output were put in place to make things easier when the application was made bash-pipeable. Or to write the output to a tarball. But now the input and output have always got to be directories.

tjungblu:
@arjunrn `os.Stdin` is also just a file to scan, the same abstractions work the exact same way. I'll add this later today. 

From your comments I would like to also split out the actual business logic from the worker into a dedicated package so it's properly testable, I don't think maintaining a a full input/output system is worth for just adding bash pipe support. It's not like there is already a working io system in the standard sdk.

pkg/omitter/fileomitter.go
tjungblu: that's actually a mistake, since it mentions it would use relative path globbing later on. I'll fix the design doc later to clarify.

pkg/traversal/reporter.go
```diff
@@ -0,0 +1,86 @@
...
+
+	s.omissions = append(s.omissions, path)
+}
+
+func (s *SimpleReporter) ReportObfuscators(obfuscators []obfuscator.Obfuscator) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	for _, o := range obfuscators {
+		s.replacements = append(s.replacements, o.Report())
```
pweil-, line 76: another advantage of just using an immutable object may be that you do not hold a lock at the mercy of "alien" code.  Ie. the duration of however long an obfuscator takes to report since it would also block other methods of this class.

Scoped locking would work here too if you aggregated replacements in the loop then appended them post loop.

pweil- (commented):
will look more if I have a chance, but Meetingpalooza is starting now for me :-) 

Thanks for tagging me, always fun!

arjunrn:
@tjungblu You cannot write a directory to `os.Stdout` to be read in through `os.Stdin`.

pkg/cli/cli_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestExistingEmptyDir(t *testing.T) {
+	testDir, err := os.MkdirTemp(os.TempDir(), "test-dir-*")
+	require.NoError(t, err)
+	defer func() {
+		_ = os.RemoveAll(testDir)
```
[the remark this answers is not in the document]
arjunrn: We have a linter to verify that error values are always checked. This is used to explicitly inform the linter that we are ignoring the error value.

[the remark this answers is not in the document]
pweil-: ack, that makes sense.  Thanks Arjun.  Don't upset the linter gods then :-D

pkg/traversal/worker.go
tjungblu: that's the whole point and hence the explanation of why there is no usage of defer for closing the handles. How would you write that in a comment?

pkg/cli/cli_test.go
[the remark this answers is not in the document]
tjungblu: yep, linters..

pkg/traversal/reporter.go
tjungblu: refactored that all the way to just report at the end without locking, thx for the suggestion.

pkg/traversal/traversal.go
tjungblu: made workerInput private, thanks. 

> Also what's the reason for passing them in here in ProcessQueue() instead of the new function?

it's to decouple the worker logic from business logic. Otherwise the worker factory would need to know that there is a work channel and some error channel. The worker is created in the cli.go now, there's no notion of channels - will make it clearer by completely pulling out all business logic from the traversal package. Then it's purpose is to only handle the channels and take the files from the traversal.

pkg/traversal/traversal.go
tjungblu: I had that failing yesterday when I ran this end2end, I'll add a test to verify for regressions

tjungblu:
@arjunrn have a look, I added some bash piping just with the business logic instead of going with an io system.

>  cat ~/Downloads/mgs/must-gather.aws/quay-io-openshift-release-dev-ocp-v4-0-art-dev-sha256-eede85cbd33388f7e18954e5ce8ad9616e91df05ace604966c0a44111d85c897/namespaces/openshift-etcd/pods/etcd-ip-10-0-187-218.ec2.internal/etcd-ip-10-0-187-218.ec2.internal.yaml | ./must-gather-clean

pkg/traversal/reporter.go
tjungblu: named the lock now after what it's guarding in the Reporters

pkg/traversal/worker.go
tjungblu: (removed the comment now)

pkg/omitter/fileomitter.go
tjungblu: fixed the design doc to make it clear the globbings are fully by path

tjungblu:
I'm also not too happy about the interfaces in the cleaner package, so if you have anything better let me know. I'm missing loads more tests there as well and we need to add proper end2end testing, I can't always run the whole thing manually

tjungblu:
added an e2e test now as well, test coverage seems somewhat improved as well:


```
?   	github.com/openshift/must-gather-clean/cmd/must-gather-clean	[no test files]
?   	github.com/openshift/must-gather-clean/pkg/cli	[no test files]
ok  	github.com/openshift/must-gather-clean/pkg/input	0.020s	coverage: 63.9% of statements
ok  	github.com/openshift/must-gather-clean/pkg/kube	0.029s	coverage: 84.2% of statements
ok  	github.com/openshift/must-gather-clean/pkg/obfuscator	0.050s	coverage: 92.9% of statements
ok  	github.com/openshift/must-gather-clean/pkg/omitter	0.029s	coverage: 96.7% of statements
ok  	github.com/openshift/must-gather-clean/pkg/output	0.030s	coverage: 79.5% of statements
ok  	github.com/openshift/must-gather-clean/pkg/schema	0.030s	coverage: 73.3% of statements
ok  	github.com/openshift/must-gather-clean/pkg/traversal	0.029s	coverage: 83.1% of statements
```

vs.

```
?   	github.com/openshift/must-gather-clean/cmd/must-gather-clean	[no test files]
ok  	github.com/openshift/must-gather-clean/pkg/cleaner	0.029s	coverage: 81.6% of statements
ok  	github.com/openshift/must-gather-clean/pkg/cli	2.822s	coverage: 57.3% of statements
ok  	github.com/openshift/must-gather-clean/pkg/kube	0.029s	coverage: 89.5% of statements
ok  	github.com/openshift/must-gather-clean/pkg/obfuscator	0.052s	coverage: 91.2% of statements
ok  	github.com/openshift/must-gather-clean/pkg/omitter	0.029s	coverage: 94.3% of statements
ok  	github.com/openshift/must-gather-clean/pkg/reporting	0.029s	coverage: 81.2% of statements
ok  	github.com/openshift/must-gather-clean/pkg/schema	0.040s	coverage: 73.3% of statements
ok  	github.com/openshift/must-gather-clean/pkg/traversal	1.028s	coverage: 81.8% of statements
```

I'll add the relpath test and I'd be good to go.

pkg/traversal/traversal.go
tjungblu: added

.gitignore
```diff
@@ -8,7 +8,10 @@
 bin/
 
 # reports created by debugging/testing
-report.yaml
+/report.yaml
+
+# cleaning result from e2e test
+pkg/cli/testfiles/release-4.8.cleaned
```
arjunrn, line 14: Should be part of e2e PR

cmd/must-gather-clean/root.go
```diff
@@ -36,24 +36,27 @@ var rootCmd = &cobra.Command{
...
+	_ = rootCmd.MarkFlagRequired("input")
+	_ = rootCmd.MarkFlagRequired("output")
+
 	fs := goflag.NewFlagSet("", goflag.ExitOnError)
 	klog.InitFlags(fs)
 	rootCmd.Flags().AddGoFlagSet(fs)
 }
 
 func main() {
+	initFlags()
```
arjunrn, line 58: This was refactored for the bash-piping right?

go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/spf13/cobra v1.1.3
 	github.com/stretchr/testify v1.7.0
 	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
```
arjunrn, line 11: We should use the v2 or v3 yaml package.

pkg/cleaner/cleaning.go
```diff
@@ -0,0 +1,138 @@
...
+	// ObfuscatePlainTextFile obfuscates a text file and writes the result into the outputFile.
+	ObfuscatePlainTextFile(inputFile string, outputFile string) error
+}
+
+type ContentObfuscator struct {
+	Obfuscator obfuscator.ReportingObfuscator
+}
+
+type FileCleaner struct {
+	ContentObfuscator
```
arjunrn, line 36: Is it really necessary to embed a struct which contains an interface?

pkg/cleaner/cleaning.go
```diff
@@ -0,0 +1,138 @@
...
+
+	inputFolder  string
+	outputFolder string
+
+	omitter omitter.ReportingOmitter
```
arjunrn, line 41: The omitter tracks what's omitted. But the FileProcessor actually omits the file. This seems wrong to me.

pkg/cleaner/cleaning.go
arjunrn: Also what's the point of using the `ReportingObfuscator` interface type here when it always has to be a `MultiObfuscator`.

pkg/cleaner/cleaning.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = os.MkdirAll(writePathParentDir, 0755)
+	if err != nil {
+		return fmt.Errorf("failed to create directory %s: %c", writePathParentDir, err)
+	}
+
+	outputOsFile, err := os.OpenFile(writePath, os.O_CREATE|os.O_WRONLY, 0666)
```
arjunrn, line 94: I assume we don't use `os.Create` here because we want it to error when the file already exists?

pkg/obfuscator/noop_obfuscator.go
```diff
@@ -0,0 +1,26 @@
+package obfuscator
+
+// NoopObfuscator is only used for testing purposes across packages. It does obfuscate nothing.
```
arjunrn, line 3: Then it should be in a test file.

pkg/obfuscator/obfuscator.go
```diff
@@ -6,6 +6,10 @@ type Obfuscator interface {
 	Path(string) string
 	// Contents takes string as input and return the obfuscated string
 	Contents(string) string
-	// Report returns a map of words and their replacements
+}
+
+type ReportingObfuscator interface {
+	Obfuscator
```
arjunrn, line 12: If everything is a `ReportingObfuscator` why do we need 2 different interfaces?

pkg/obfuscator/tracker.go
```diff
@@ -8,9 +8,9 @@ import (
 
 type GenerateReplacement func(string) string
 
-// ReplacementTracker is used to track and generate replacements used by obfuscators
+// ReplacementTracker is used to track and generate Replacements used by obfuscators
```
arjunrn, line 11: Refactoring glitch.

pkg/obfuscator/tracker.go
```diff
@@ -8,9 +8,9 @@ import (
...
 type ReplacementTracker interface {
-	// Initialize initializes the tracker with some existing replacements. It should be called only once and before
+	// Initialize initializes the tracker with some existing Replacements. It should be called only once and before
```
arjunrn, line 13: Refactoring glitch.

pkg/obfuscator/tracker.go
```diff
@@ -72,7 +72,7 @@ func (s *SimpleTracker) Initialize(replacements map[string]string) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 	if len(s.mapping) > 0 {
-		klog.Exitf("tracker was initialized more than once or after some replacements were already added.")
+		klog.Exitf("tracker was initialized more than once or after some Replacements were already added.")
```
arjunrn, line 75: Refactoring glitch.

pkg/cleaner/cleaning.go
tjungblu: you have used OpenFile here: https://github.com/openshift/must-gather-clean/blob/24c96bc6cd283ace209a1f22e45d0f692f508816/pkg/output/output.go#L39

Appending is certainly not what we want, is it?

pkg/obfuscator/obfuscator.go
tjungblu: for logic that doesn't need reporting?

pkg/cleaner/cleaning.go
tjungblu: it was how the logic was before and not the goal of the refactoring. 
AFAIR you've added that interface with that behaviour :)

pkg/cleaner/cleaning.go
tjungblu: yeah that's exactly the place that I'm not happy about in the abstraction. I want to split the logic of obfuscating content and files. 

There's also no single place that requires a MultIObfuscator as the struct, the cleaner works on the ReportingObfuscator.

pkg/obfuscator/noop_obfuscator.go
tjungblu: I thought it would work, but if it's a test file it can't be exported, this is needed in both cleaner and reporting packages.

pkg/cleaner/cleaning.go
tjungblu: > yeah that's exactly the place that I'm not happy about in the abstraction. I want to split the logic of obfuscating content and files.

it's probably clearer with the bash piping, since the ContentObfuscator is used there to obfuscate stdin. I don't want to duplicate the content obfuscation logic with the scanner once more in the fileCleaner. I'd prefer to have less duplication here - what's the problem with having a single value struct embedded?

pkg/cleaner/cleaning.go
arjunrn: No, the `Omiter` didn't have any reporting functionality for the reason mentioned above.

pkg/cleaner/cleaning.go
arjunrn: Yep, the current logic is not correct.

pkg/obfuscator/noop_obfuscator.go
arjunrn: In that case copy it. Having test code in non-test files seems wrong.

pkg/obfuscator/obfuscator.go
arjunrn: Because the way I see it there are different types of obfuscators and they obfuscate stuff and also report what they did.

pkg/obfuscator/multi_obfuscator.go
```diff
@@ -0,0 +1,46 @@
+package obfuscator
+
+type MultiObfuscator struct {
+	obfuscators []ReportingObfuscator
+}
+
+func (m *MultiObfuscator) Path(s string) string {
```
sairameshv, line 7: What is the advantage here to return a single string after the multiple obfuscations?

pkg/omitter/reporting_omitter.go
```diff
@@ -0,0 +1,73 @@
...
+
+type MultiReportingOmitter struct {
+	fileOmitters []FileOmitter
+	k8sOmitters  []KubernetesResourceOmitter
+
+	omittedPathsLock sync.Mutex
+	omittedPaths     []string
+}
+
+func (m *MultiReportingOmitter) OmitPath(path string) (bool, error) {
```
sairameshv, line 17: Please add a comment/description on what this function does.

pkg/omitter/reporting_omitter.go
```diff
@@ -0,0 +1,73 @@
...
+func (m *MultiReportingOmitter) Report() []string {
+	m.omittedPathsLock.Lock()
+	defer m.omittedPathsLock.Unlock()
+
+	var copySlice []string
+	for i := 0; i < len(m.omittedPaths); i++ {
+		copySlice = append(copySlice, m.omittedPaths[i])
+	}
+
+	return copySlice
```
sairameshv, line 56: Why can't we just return `m.omittedPaths` ?
The intention here is to improve the memory efficiency (The new slice memory gets allocated on a heap as the same is getting returned)

pkg/traversal/traversal.go
```diff
@@ -1,59 +1,37 @@
...
+	w.workers = make([]QueueProcessor, w.workerCount)
 	for i := 0; i < w.workerCount; i++ {
-		wk := newWorker(i+1, w.obfuscators, w.fileOmitters, w.k8sOmitters, queue, w.writer, errorCh)
-		w.workers[i] = wk
+		w.workers[i] = w.workerFactory(i + 1)
 		wg.Add(1)
-		go func() {
-			wk.run()
+		go func(i int, queue chan workerInput, errorCh chan error) {
+			w.workers[i].ProcessQueue(queue, errorCh)
```
sairameshv, line 32: Perfect :)

pkg/cleaner/cleaning.go
```diff
@@ -0,0 +1,155 @@
...
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/openshift/must-gather-clean/pkg/kube"
+	"github.com/openshift/must-gather-clean/pkg/obfuscator"
+	"github.com/openshift/must-gather-clean/pkg/omitter"
+)
+
+type Processor interface {
```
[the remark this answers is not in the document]
tjungblu: added

pkg/obfuscator/multi_obfuscator.go
tjungblu: this is the business logic as described in the design doc: https://docs.google.com/document/d/1IDUfaw2IOPo7fRK142rsiAcxwl7tAHXJwLhaoyhambI/edit#

you can also read more about the behaviour in the readme: https://github.com/openshift/must-gather-clean#chaining-obfuscators-and-side-effects

pkg/omitter/reporting_omitter.go
tjungblu: this is documented on the interface: https://github.com/openshift/must-gather-clean/pull/43/files/dbc7e3c73711cdada169ef2c15dcf9887b97af46#diff-2ba966afd5b579a31a3e3e44617e9a44c92eb64e1f58a8d97ddcaaa546398764R8

`	// OmitPath takes the relative path of the file and its return indicates if the file should be omitted.`

pkg/omitter/reporting_omitter.go
tjungblu: this is a defensive copy, so a caller is not able to: 
1. change its contents (eg resize on grow)
2. reading can be done outside of the locking

Imagine another thread changing the omittedPathsLock while another thread holds on to the slice internally referenced.

PR #44, opened 2021-09-02
consistent MAC address obfuscation

sairameshv:
Code changes to support the consistent MAC address obfuscations.
Also refactored some of the code to make it generic for the obfuscation of both the IP as well as MAC addresses.


pkg/obfuscator/common.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"fmt"
+
+	"k8s.io/klog/v2"
+)
+
+// obfsGenerator consists of the required fields for the consistent,static obfuscations and the count of the obfuscations
+// This implements the methods static, consistent inorder to return the required replacement based on the replacementType
+type obfsGenerator struct {
```
tjungblu, line 11: I'd put that into a generator.go file instead

tjungblu: and I guess you can also name it that way, package is already called obfuscator

pkg/obfuscator/mac_address.go
```diff
@@ -19,8 +29,15 @@ func (m *macAddressObfuscator) Path(s string) string {
...
 	for _, match := range matches {
-		s = strings.Replace(s, match, StaticMacReplacement, -1)
-		m.ReplacementTracker.AddReplacement(match, StaticMacReplacement)
+		var replacement string
+		switch m.replacementType {
+		case schema.ObfuscateReplacementTypeStatic:
+			replacement = m.GenerateIfAbsent(match, match, m.obfsGenerator.generateStaticReplacement)
+		case schema.ObfuscateReplacementTypeConsistent:
+			replacement = m.GenerateIfAbsent(match, match, m.obfsGenerator.generateConsistentReplacement)
+		}
```
tjungblu, line 38: the whole switch looks also very similar to the one in the ip obfuscator, do you think we can refactor that too? I guess it would look the same for the domain one too.

pkg/obfuscator/mac_address.go
```diff
@@ -29,14 +46,23 @@ func (m *macAddressObfuscator) Report() map[string]string {
...
+func NewMacAddressObfuscator(replacementType schema.ObfuscateReplacementType) (Obfuscator, error) {
+	if replacementType != schema.ObfuscateReplacementTypeStatic && replacementType != schema.ObfuscateReplacementTypeConsistent {
+		return nil, fmt.Errorf("unsupported replacement type: %s", replacementType)
+	}
 	// this regex differs from the standard `(?:[0-9a-fA-F]([:-])?){12}`, to not match very frequently happening UUIDs in K8s
 	// the main culprit is the support for squashed MACs like '69806FE67C05', which won't be supported with the below
 	regex := regexp.MustCompile(`([0-9a-fA-F]{2}[:-]){5}[0-9a-fA-F]{2}`)
 
 	reporter := NewSimpleTracker()
+	generator := obfsGenerator{
```
tjungblu, line 58: can you wrap this with a New* function as well? that way you can decouple the struct and its interface

pkg/obfuscator/mac_address_test.go
```diff
@@ -4,28 +4,36 @@ import (
 	"fmt"
```
tjungblu, line 4: tests look great, thanks!

pkg/obfuscator/common.go
sairameshv: Sure.

pkg/obfuscator/mac_address.go
sairameshv: True, addressed

pkg/obfuscator/generator.go
```diff
@@ -0,0 +1,33 @@
...
+	r := fmt.Sprintf(g.template, g.count)
+	return r
+}
+
+func (g *generator) generateStaticReplacement(_ string) string {
+	return g.static
+}
+
+// NewGenerator creates a generator objects and populates with the provided arguments
+func NewGenerator(template, static string, count int) *generator {
```
arjunrn, line 31: An exported function should not return an unexported type.

pkg/obfuscator/mac_address.go
```diff
@@ -19,8 +29,15 @@ func (m *macAddressObfuscator) Path(s string) string {
...
-		s = strings.Replace(s, match, StaticMacReplacement, -1)
-		m.ReplacementTracker.AddReplacement(match, StaticMacReplacement)
+		var replacement string
+		switch m.replacementType {
+		case schema.ObfuscateReplacementTypeStatic:
+			replacement = m.GenerateIfAbsent(match, match, m.obfsGenerator.generateStaticReplacement)
+		case schema.ObfuscateReplacementTypeConsistent:
+			replacement = m.GenerateIfAbsent(match, match, m.obfsGenerator.generateConsistentReplacement)
+		}
+		s = strings.Replace(s, match, replacement, -1)
```
arjunrn, line 39: `strings.ReplaceAll()`

pkg/obfuscator/generator.go
arjunrn: Since `count` will always start from 0 other than the tests it should not be passed as a parameter.

arjunrn: You don't need to initialize it either because that's the [default value](https://tour.golang.org/basics/12).

pkg/obfuscator/mac_address.go
```diff
@@ -1,15 +1,25 @@
...
+
+	"github.com/openshift/must-gather-clean/pkg/schema"
 )
 
-const StaticMacReplacement = "xx:xx:xx:xx:xx:xx"
+const (
+	// staticMacReplacement refers to a static replacement for any identified MAC address.
+	staticMacReplacement = "xx:xx:xx:xx:xx:xx"
+	// consistentMACTemplate refers to a consistent replacement for any identified MAC address
+	consistentMACTemplate = "x-mac-%06d-x"
```
arjunrn, line 15: Try to make this around as many characters as there are in a MAC address.

pkg/obfuscator/mac_address_test.go
```diff
@@ -34,17 +42,72 @@ func TestMacReplacementSuper(t *testing.T) {
...
 	} {
 		t.Run(tc.name, func(t *testing.T) {
-			o := NewMacAddressObfuscator()
+			o, _ := NewMacAddressObfuscator(schema.ObfuscateReplacementTypeStatic)
 			assert.Equal(t, tc.expectedOutput, o.Contents(tc.input))
 		})
 	}
 }
+
+func TestMACConsistentObfuscator(t *testing.T) {
```
arjunrn, line 60: Add a test for the same mac address with dashes and colons.

pkg/obfuscator/generator.go
sairameshv: Yeah, I will address these

pkg/obfuscator/ip_test.go
```diff
@@ -207,6 +206,14 @@ func TestIPObfuscatorConsistent(t *testing.T) {
 				"192.168.1.34": "x-ipv4-000010-x",
 			},
 		},
+		{
+			name:   "standard ipv4 with colons and standard ipv4 with dashes between should map to the same obfuscation value",
+			input:  []string{"obfuscate 10.0.129.220 and 10-0-129-220"},
+			output: []string{"obfuscate x-ipv4-000001-x and x-ipv4-000001-x"},
+			report: map[string]string{
+				"10.0.129.220": "x-ipv4-000001-x",
```
tjungblu, line 214: thanks again for catching that one, that even went pass my end2end tests the other day

pkg/obfuscator/tracker.go
```diff
@@ -54,18 +54,23 @@ func (s *SimpleTracker) AddReplacement(original string, replacement string) {
 	s.mapping[original] = replacement
 }
 
-func (s *SimpleTracker) GenerateIfAbsent(key string, generator GenerateReplacement) string {
+func (s *SimpleTracker) GenerateIfAbsent(key string, generator GenerateReplacement) (string, bool) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
+	// returning the empty string if the replacement is already present
+	// GenerateIfAbsent only should generate the replacement if the key is a new one
+	// This helps in avoiding the addition of case-sensitive alternatives to the report
```
tjungblu, line 62: why is that a problem? I don't it it makes sense to add such a return value here. Can you explain why this is necessary instead of making AddReplacement idempotent?

tjungblu: plus, I think that way you are also circumventing an important check that we don't add the same key with a different value multiple times.

pkg/obfuscator/mac_address.go
sairameshv: Addressed

pkg/obfuscator/mac_address_test.go
sairameshv: Done

pkg/obfuscator/tracker.go
sairameshv: Lets say the same MAC address has been observed twice, one containing the upper case and another with lower case alphabets. Ex: "2C:54:91:88:C9:E3 can also be displayed as 2c:54:91:88:c9:e3"
When we call the Generate function, we normalize the key string to the uppercase and call the function which is "2C:54:91:88:C9:E3"
Though the substituted value remains the same("xxx-mac-000001-xxx"), both key-value replacement pairs would be appended to the report mapping which I feel is ambiguous. 
{
"2C:54:91:88:C9:E3": "xxx-mac-000001-xxx",
"2c:54:91:88:c9:e3": "xxx-mac-000001-xxx"
}
Hence, I have given a thorough thought and felt this extra return (a boolean) helps in verifying if the data is already added to the report and if not, can be added.

PR #46, opened 2021-09-07
Source git version from build-machinery-go Makefile

josefkarasek:
[Makefile](https://github.com/openshift/build-machinery-go/blob/2b271bb3a0ad466045cd6da5c9423084e9cf68f0/make/lib/golang.mk#L63-L68) in build-machinery-go offers git metadata that can be included in the `version` sub-command.


hack/bump-version.sh
```diff
@@ -1,31 +0,0 @@
-#!/bin/bash
```
josefkarasek, line 1: What is the purpose of this file? 
Did CFE or somebody else provide it?
Is is part of some existing CI?

tjungblu: we copied and pasted everything from odo: https://github.com/openshift/odo

but cool support for the versioning, how do we do the release version? via git tag?

cmd/must-gather-clean/version.go
```diff
@@ -1,21 +1,16 @@
...
-const (
-	VERSION = "v0.0.1"
-)
-
 // versionCmd represents the version command
 var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "Print the version of the tool",
 	Run: func(_ *cobra.Command, _ []string) {
-		fmt.Println(VERSION)
```
josefkarasek, line 18: Example output:
```
Version: pkg.Version{Version:"v0.0.1-0-g6ce19dc", GitCommit:"6ce19dc", BuildDate:"2021-09-07T12:08:00Z", GoOs:"linux", GoArch:"amd64"}
```
`v0.0.1-0-g6ce19dc`
* tag v0.0.1
* 0 commits ahead of v0.0.1
* commit SHA

josefkarasek: https://git-scm.com/docs/git-describe#_examples

hack/bump-version.sh
josefkarasek: Yes, it comes from a [tag](https://github.com/openshift/build-machinery-go/blob/2b271bb3a0ad466045cd6da5c9423084e9cf68f0/make/lib/golang.mk#L53): `git describe --long --tags --abbrev=7 --match 'v[0-9]*'`

tjungblu:
good news, we can release as a community tool through GH release. So we can take this approach for now :)

tjungblu (approved):
/lgtm

PR #48, opened 2021-09-08
Support for Static Domain Obfuscation

sairameshv:
Code changes to support the static domain obfuscation.

### DIFF
--- a/pkg/obfuscator/domain.go
+++ b/pkg/obfuscator/domain.go
@@ -4,18 +4,22 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+
+	"github.com/openshift/must-gather-clean/pkg/schema"
 )
 
 const (
-	domainPattern      = `([a-zA-Z0-9\.]*\.)?(%s)`
-	obfuscatedTemplate = "domain%07d"
+	domainPattern           = `([a-zA-Z0-9\.]*\.)?(%s)`
+	obfuscatedTemplate      = "domain%07d"
+	staticDomainReplacement = "example-domain.com"
 )
 
 type domainObfuscator struct {
 	ReplacementTracker
-	domainCount    int
-	domainPatterns []*regexp.Regexp
-	domainMapping  map[string]string
+	replacementType schema.ObfuscateReplacementType
+	domainPatterns  []*regexp.Regexp
+	domainMapping   map[string]string
+	obfsGenerator   generator
 }
 
 func (d *domainObfuscator) Path(s string) string {
@@ -54,13 +58,18 @@ func (d *domainObfuscator) obfuscatedDomain(domain string) string {
 	if replacement, ok := d.domainMapping[domain]; ok {
 		return replacement
 	}
-	d.domainCount++
-	replacement := fmt.Sprintf(obfuscatedTemplate, d.domainCount)
+	var replacement string
+	switch d.replacementType {
+	case schema.ObfuscateReplacementTypeStatic:
+		replacement = d.obfsGenerator.generateStaticReplacement()
+	case schema.ObfuscateReplacementTypeConsistent:
+		replacement = d.obfsGenerator.generateConsistentReplacement()
+	}
 	d.domainMapping[domain] = replacement
 	return replacement
 }
 
-func NewDomainObfuscator(domains []string) (Obfuscator, error) {
+func NewDomainObfuscator(domains []string, replacementType schema.ObfuscateReplacementType) (Obfuscator, error) {
 	patterns := make([]*regexp.Regexp, len(domains))
 	for i, d := range domains {
 		dd := strings.ReplaceAll(d, ".", "\\.")
@@ -70,9 +79,13 @@ func NewDomainObfuscator(domains []string) (Obfuscator, error) {
 		}
 		patterns[i] = p
 	}
+	// creating a new generator object
+	generator := newGenerator(obfuscatedTemplate, staticDomainReplacement)
 	return &domainObfuscator{
 		ReplacementTracker: NewSimpleTracker(),
 		domainPatterns:     patterns,
 		domainMapping:      map[string]string{},
+		replacementType:    replacementType,
+		obfsGenerator:      *generator,
 	}, nil
 }
--- a/pkg/obfuscator/domain_test.go
+++ b/pkg/obfuscator/domain_test.go
@@ -3,9 +3,10 @@ package obfuscator
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/openshift/must-gather-clean/pkg/schema"
 )
 
 func TestDomainObfuscatorContents(t *testing.T) {
@@ -80,7 +81,7 @@ func TestDomainObfuscatorContents(t *testing.T) {
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
-			o, err := NewDomainObfuscator(tc.domains)
+			o, err := NewDomainObfuscator(tc.domains, schema.ObfuscateReplacementTypeConsistent)
 			require.NoError(t, err)
 			for idx, i := range tc.input {
 				output := o.Contents(i)
@@ -117,7 +118,7 @@ func TestDomainObfuscator_FileName(t *testing.T) {
 		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
-			o, err := NewDomainObfuscator(tc.domains)
+			o, err := NewDomainObfuscator(tc.domains, schema.ObfuscateReplacementTypeConsistent)
 			require.NoError(t, err)
 			output := o.Path(tc.input)
 			assert.Equal(t, tc.output, output)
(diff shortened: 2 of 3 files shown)


pkg/obfuscator/domain.go
```diff
@@ -54,13 +58,18 @@ func (d *domainObfuscator) obfuscatedDomain(domain string) string {
...
 	}
-	d.domainCount++
-	replacement := fmt.Sprintf(obfuscatedTemplate, d.domainCount)
+	var replacement string
+	switch d.replacementType {
+	case schema.ObfuscateReplacementTypeStatic:
+		replacement = d.obfsGenerator.generateStaticReplacement()
+	case schema.ObfuscateReplacementTypeConsistent:
+		replacement = d.obfsGenerator.generateConsistentReplacement()
+	}
```
tjungblu, line 67: it's now the third time I find that exact same switch statement, can we pull that into a dedicated struct+interface along with the replacementType as input and refactor the other obfuscators towards that as well?

pkg/obfuscator/domain.go
```diff
@@ -4,18 +4,22 @@ import (
...
+
+	"github.com/openshift/must-gather-clean/pkg/schema"
 )
 
 const (
-	domainPattern      = `([a-zA-Z0-9\.]*\.)?(%s)`
-	obfuscatedTemplate = "domain%07d"
+	domainPattern           = `([a-zA-Z0-9\.]*\.)?(%s)`
+	obfuscatedTemplate      = "domain%07d"
+	staticDomainReplacement = "example-domain.com"
```
tjungblu, line 14: shall we just generate `x`.repeat(n)+ ".com"? I think example-domain might be a bit confusing

pkg/obfuscator/domain.go
```diff
@@ -54,13 +58,18 @@ func (d *domainObfuscator) obfuscatedDomain(domain string) string {
 	if replacement, ok := d.domainMapping[domain]; ok {
```
tjungblu, line 58: the whole thing suffers from race conditions now because the domainMapping is not locked: it can happen that one thread has a domain mapping replaced with another mapping.

we have to do the same generateIfAbsent that we do in the other obfuscators too. As mentioned below, it makes sense to just refactor what we already did in MAC+IP into dedicated and correct logical structure and then reuse it here.

tjungblu: actually, it suffered from races before as well :(

tjungblu (changes requested):

pkg/obfuscator/domain.go
sairameshv: Yeah, I will address this.

pkg/obfuscator/domain.go
sairameshv: As we are referring to the static replacement, what is the significance of "n" here?

pkg/obfuscator/domain.go
```diff
@@ -44,23 +47,17 @@ func (d *domainObfuscator) replaceDomains(input string) string {
 				replacement = obfuscatedBaseDomain
 			}
 			output = strings.ReplaceAll(output, m[0], replacement)
-			d.AddReplacement(m[0], replacement)
 		}
 	}
 	return output
 }
 
 func (d *domainObfuscator) obfuscatedDomain(domain string) string {
```
tjungblu, line 55: I think we can remove the method and just directly call the obfsGenerator

pkg/obfuscator/domain.go
```diff
@@ -4,18 +4,21 @@ import (
...
+	obfuscatedTemplate      = "domain%07d"
+	staticDomainReplacement = "obfuscated.com"
 )
 
 type domainObfuscator struct {
 	ReplacementTracker
-	domainCount    int
-	domainPatterns []*regexp.Regexp
-	domainMapping  map[string]string
+	replacementType schema.ObfuscateReplacementType
```
tjungblu, line 19: why do we need that here? that should be in the generator

pkg/obfuscator/domain.go
```diff
@@ -44,23 +47,17 @@ func (d *domainObfuscator) replaceDomains(input string) string {
...
 }
 
 func (d *domainObfuscator) obfuscatedDomain(domain string) string {
-	if replacement, ok := d.domainMapping[domain]; ok {
-		return replacement
-	}
-	d.domainCount++
-	replacement := fmt.Sprintf(obfuscatedTemplate, d.domainCount)
-	d.domainMapping[domain] = replacement
+	replacement := d.obfsGenerator.generateReplacement(d.replacementType, domain, d.ReplacementTracker)
```
tjungblu, line 56: `d.replacementType` shouldn't be needed anymore

pkg/obfuscator/domain_test.go
```diff
@@ -127,7 +125,60 @@ func TestDomainObfuscator_FileName(t *testing.T) {
...
+		},
+		{
+			name:    "non-matching domain",
+			domains: []string{"test.com"},
+			input:   []string{"report.test"},
+			output:  []string{"report.test"},
+			report:  map[string]string{},
+		},
+		{
+			name:    "Multiple-Matching Domains",
```
tjungblu, line 159: I think you rather want to test whether in the same input string, you're catching multiple domains? 
so maybe input: "The first domain is report.test.com and the second domain is example-test.com"

otherwise you just test that golang for loops work

pkg/obfuscator/generator.go
```diff
@@ -27,7 +29,22 @@ func (g *generator) generateStaticReplacement() string {
 	return g.static
 }
 
+// generateReplacement returns the replacement based on the replacementType argument
+func (g *generator) generateReplacement(replacementType schema.ObfuscateReplacementType, key string, tracker ReplacementTracker) string {
```
tjungblu, line 33: parameter is not necessary anymore, you can reference from `g.replacementType`

pkg/obfuscator/generator.go
```diff
@@ -27,7 +29,22 @@ func (g *generator) generateStaticReplacement() string {
...
+		replacement = tracker.GenerateIfAbsent(key, g.generateConsistentReplacement)
+	}
+	return replacement
+}
+
 // newGenerator creates a generator objects and populates with the provided arguments
-func newGenerator(template, static string) *generator {
-	return &generator{template: template, static: static}
+func newGenerator(template, static string, replacementType schema.ObfuscateReplacementType) (*generator, error) {
+	if replacementType != schema.ObfuscateReplacementTypeStatic && replacementType != schema.ObfuscateReplacementTypeConsistent {
```
tjungblu, line 46: can you add a test for that statement? Try passing in some other string, I think you need to do a type cast to schema.ObfuscateReplacementType - which is just a string anyway

pkg/obfuscator/ip.go
```diff
@@ -80,8 +74,8 @@ func NewIPObfuscator(replacementType schema.ObfuscateReplacementType) (Obfuscato
 	return &ipObfuscator{
 		ReplacementTracker: NewSimpleTracker(),
 		replacements: map[*regexp.Regexp]*generator{
-			ipv4Pattern: {template: consistentIPv4Template, static: obfuscatedStaticIPv4},
-			ipv6Pattern: {template: consistentIPv6Template, static: obfuscatedStaticIPv6},
+			ipv4Pattern: {template: consistentIPv4Template, static: obfuscatedStaticIPv4, replacementType: replacementType},
```
tjungblu, line 77: can't comment on the above line, but you can remove the `if replacementType != ...` checks now

pkg/obfuscator/mac_address.go
```diff
@@ -58,7 +52,10 @@ func NewMacAddressObfuscator(replacementType schema.ObfuscateReplacementType) (O
```
tjungblu, line 52: same as the IP obfuscator: can't comment on the above line, but you can remove the if replacementType != ... checks now

pkg/obfuscator/ip.go
```diff
@@ -58,13 +58,7 @@ func (o *ipObfuscator) replace(s string) string {
...
 			cleaned := strings.ReplaceAll(m, "-", ".")
 			if ip := net.ParseIP(cleaned); ip != nil {
-				var replacement string
-				switch o.replacementType {
-				case schema.ObfuscateReplacementTypeStatic:
-					replacement = o.GenerateIfAbsent(cleaned, gen.generateStaticReplacement)
-				case schema.ObfuscateReplacementTypeConsistent:
-					replacement = o.GenerateIfAbsent(cleaned, gen.generateConsistentReplacement)
-				}
+				replacement := gen.generateReplacement(o.replacementType, cleaned, o.ReplacementTracker)
```
tjungblu, line 61: replacement type not necessary and can be removed from the ipObfuscator struct

pkg/obfuscator/mac_address.go
```diff
@@ -31,13 +31,7 @@ func (m *macAddressObfuscator) Contents(s string) string {
...
 		// normalizing the MAC Address string to the Uppercase so as to avoid the duplicate reporting
 		match := strings.ToUpper(strings.ReplaceAll(mac, "-", ":"))
-		var replacement string
-		switch m.replacementType {
-		case schema.ObfuscateReplacementTypeStatic:
-			replacement = m.GenerateIfAbsent(match, m.obfsGenerator.generateStaticReplacement)
-		case schema.ObfuscateReplacementTypeConsistent:
-			replacement = m.GenerateIfAbsent(match, m.obfsGenerator.generateConsistentReplacement)
-		}
+		replacement := m.obfsGenerator.generateReplacement(m.replacementType, match, m.ReplacementTracker)
```
tjungblu, line 34: replacement type not necessary and can be removed from the macAddressObfuscator struct

pkg/obfuscator/domain.go
sairameshv: Addressed

pkg/obfuscator/domain.go
sairameshv: Addressed

pkg/obfuscator/domain_test.go
sairameshv: Addressed

pkg/obfuscator/generator.go
sairameshv: Done.

pkg/obfuscator/domain.go
```diff
@@ -4,18 +4,21 @@ import (
...
+	"github.com/openshift/must-gather-clean/pkg/schema"
 )
 
 const (
-	domainPattern      = `([a-zA-Z0-9\.]*\.)?(%s)`
-	obfuscatedTemplate = "domain%07d"
+	domainPattern                      = `([a-zA-Z0-9\.]*\.)?(%s)`
+	obfuscatedTemplate                 = "domain%07d"
+	staticDomainReplacement            = "obfuscated.com"
+	maximumSupportedObfuscationDomains = 9999999999
```
tjungblu, line 15: I think you have a couple 9's too many: 9,999,999,999. Your format only allows for seven digits.

pkg/obfuscator/generator_test.go
```diff
@@ -1,25 +1,33 @@
...
+	"fmt"
 	"testing"
 
+	"github.com/openshift/must-gather-clean/pkg/schema"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestGeneratorHappyPath(t *testing.T) {
-	g := newGenerator("%d", "x", 10)
+	g, _ := newGenerator("%d", "x", 10, schema.ObfuscateReplacementTypeStatic)
```
tjungblu, line 12: rather assign the error and then do a require:

```
g, err := newGenerator("%d", "x", 10, schema.ObfuscateReplacementTypeStatic)
require.NoError(t, err) // or so, checkout the other tests!
```

tjungblu:
good to go, thanks @sairameshv  :rocket: 

/lgtm

tjungblu (approved):

PR #49, opened 2021-09-08
omission documentation

tjungblu:


README.md
```diff
@@ -214,5 +228,73 @@ Running the above obfuscation on the string `a wonderful evening to go dancing`
...
+```
+config:
+  omit:
+  - type: Kubernetes
+    kubernetesResource:
+       kind: "Secret"
+       namespaces: ["kube-system", "default"]
+```
+
+which will only omit the secrets in the defined namespaces.
```
snarayan-redhat, line 283: If I understand the technicality correctly, doesn't it mean this?
```suggestion
which will omit only the secrets in the defined namespaces.
```
You can correct if my understanding is right.

tjungblu: absolutely correct, let me commit this

snarayan-redhat (approved):

PR #50, opened 2021-09-09
increase ip formatting limits

tjungblu:
just had a must gather where we ran over the 1m IPs in the formatting, added a fix for that and some additional testing


pkg/obfuscator/mac_address.go
```diff
@@ -58,7 +58,7 @@ func NewMacAddressObfuscator(replacementType schema.ObfuscateReplacementType) (O
 
 	reporter := NewSimpleTracker()
 	// creating a new generator object
-	generator := newGenerator(consistentMACTemplate, staticMacReplacement)
+	generator := newGenerator(consistentMACTemplate, staticMacReplacement, 9999999999)
```
sairameshv, line 61: "maximumSupportedObfuscations" can be used in order to avoid hard-coding

tjungblu: split it out into another constant to not accidentally mix things up

PR #55, opened 2021-09-14
add support for bash pipes

tjungblu:


cmd/must-gather-clean/root.go
```diff
@@ -29,31 +31,48 @@ var rootCmd = &cobra.Command{
...
+	}
+
 	fs := goflag.NewFlagSet("", goflag.ExitOnError)
 	klog.InitFlags(fs)
 	rootCmd.Flags().AddGoFlagSet(fs)
 }
 
 func main() {
+	stat, _ := os.Stdin.Stat()
+	if (stat.Mode() & os.ModeCharDevice) == 0 {
```
josefkarasek, line 70: I don't know the details, just want to make sure bitwise AND is on purpose here?

tjungblu: yep: https://stackoverflow.com/questions/43947363/detect-if-a-command-is-piped-or-not

PR #56, opened 2021-09-14
must-gather reports for gcp,asws,azure,vsphere,metal-ipv6

josefkarasek:


test/files/extract.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
josefkarasek, line 1: example script how files in this dir can be extracted and manipulated

PR #58, opened 2021-09-14
ip regex range limit

tjungblu:
that should mitigate some of the edge conditions we saw in #57 and [CFE-125](https://issues.redhat.com/browse/CFE-125)


josefkarasek:
Good stuff!

/lgtm

PR #60, opened 2021-09-15
Fix for a creation of new file incase of a static obfuscation scenarios

sairameshv:


pkg/cleaner/cleaning.go
```diff
@@ -105,18 +108,31 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
 	// we need to assess whether the file exists already to ensure we don't overwrite existing obfuscated data.
 	// that can happen while obfuscating file names and their paths.
 	// Additionally, the stat check is required because os.O_CREATE will implicitly os.O_TRUNC if a file already exist
+	// Also defining a lock to avoid collisions while creation of files in a multi-threaded environment
+	var mutexLock sync.Mutex
```
tjungblu, line 112: this will create a mutex per function call, that will not really lock anything. I think you have to add this to the struct.

And please give it a more descriptive name, maybe "pathCollisionMutex"?

pkg/cleaner/cleaning.go
```diff
@@ -88,6 +90,7 @@ func (c *FileProcessor) Process(path string) error {
 }
 
 func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile string) error {
+	var fileExt int
```
tjungblu, line 93: why define that at the top? you only use it in the error scope below

pkg/cleaner/cleaning.go
```diff
@@ -105,18 +108,31 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
 	// we need to assess whether the file exists already to ensure we don't overwrite existing obfuscated data.
 	// that can happen while obfuscating file names and their paths.
 	// Additionally, the stat check is required because os.O_CREATE will implicitly os.O_TRUNC if a file already exist
+	// Also defining a lock to avoid collisions while creation of files in a multi-threaded environment
+	var mutexLock sync.Mutex
+	mutexLock.Lock()
```
tjungblu, line 113: always accompany your lock with a defer unlock. In your code when we return through an error it would keep the lock hold forever.

you can also scope locks through defers when you wrap them in functions:
https://kylewbanks.com/blog/when-to-use-defer-in-go

which I would suggest in this case, you are passing a writePath and you return a file that does not collide.

pkg/cleaner/cleaning.go
```diff
@@ -105,18 +108,31 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
...
 	if err != nil && !os.IsNotExist(err) {
 		return fmt.Errorf("failed to determine if %s already exists: %w", writePath, err)
 	}
 	if err == nil {
-		return fmt.Errorf("file %s already exists, check whether the obfuscators overwrite each other", writePath)
+		for {
+			fileExt++
+			samplePath := writePath + "." + strconv.Itoa(fileExt)
+			_, err = os.Stat(samplePath)
+			if err != nil && os.IsNotExist(err) {
```
tjungblu, line 123: what if err != nil and it is NOT `os.IsNotExist(err)` - so a "real" error?

pkg/cleaner/cleaning.go
```diff
@@ -105,18 +108,31 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
...
+				writePath = samplePath
+				break
+			}
+			// return fmt.Errorf("file %s already exists, check whether the obfuscators overwrite each other", writePath)
```
tjungblu, line 127: please don't comment code, either it's not used then remove it or uncomment it.

pkg/cleaner/cleaning_test.go
```diff
@@ -90,7 +89,7 @@ func TestObfuscateReaderIOErrorPropagates(t *testing.T) {
 	require.ErrorIs(t, err, UnwritableErr)
 }
 
-func TestObfuscateFileOutputExistsFails(t *testing.T) {
+func TestObfuscateFileOutputExists(t *testing.T) {
```
tjungblu, line 92: can you add another test that will check that the files are created with the right pattern (ascending numbers appended)?

pkg/cleaner/cleaning.go
sairameshv: The scope of "fileExt" would be alive in case of calling this method by the same caller from multiple threads.
Thus, I felt, the number of checks, to verify if a file exists, would reduce.
Ex: 
Caller-1 called this method from Go-routine-1, Let's say fileExt has become 3
Caller-2 called this method from Go-routine-2, it can start checking the existence of file from fileExt = 4 i.e "xxx-xxx-xxx-xxx.yaml.4"

Anyway, the scope can also be limited to the error scenario only.

pkg/cleaner/cleaning.go
sairameshv: Addressed

pkg/cleaner/cleaning.go
sairameshv: Addressed

pkg/cleaner/cleaning_test.go
sairameshv: Sure

pkg/cleaner/cleaning.go
tjungblu: each goroutine has their own stack, so for each goroutine this variable will be 0 from the start anyway and they will each individually increment until they find a non-conflicting extension.

pkg/cleaner/cleaning.go
sairameshv: Addressed

pkg/cleaner/cleaning.go
```diff
@@ -105,12 +109,26 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
 	// we need to assess whether the file exists already to ensure we don't overwrite existing obfuscated data.
 	// that can happen while obfuscating file names and their paths.
 	// Additionally, the stat check is required because os.O_CREATE will implicitly os.O_TRUNC if a file already exist
+	c.pathCollisionMutex.Lock()
+	defer c.pathCollisionMutex.Unlock()
```
tjungblu, line 113: this is great, the problem is that this now locks the whole obfuscation process below as well (which is what we want to actually parallelize). 

Can you pull the locking, your added logic + OpenFile into a dedicated method that is locked? that way the lock has a clear and minimal scope

pkg/cleaner/cleaning.go
```diff
@@ -105,12 +109,26 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
...
+
 	_, err = os.Stat(writePath)
 	if err != nil && !os.IsNotExist(err) {
 		return fmt.Errorf("failed to determine if %s already exists: %w", writePath, err)
 	}
 	if err == nil {
-		return fmt.Errorf("file %s already exists, check whether the obfuscators overwrite each other", writePath)
+		var fileExt int
```
tjungblu, line 120: why not just `fileExt:=0`?

pkg/cleaner/cleaning.go
tjungblu: you can write this a bit easier to read:

if err != nil {
  if err is not exists {
    break
  } else {
    return fmt.Error with err
  }
}

pkg/cleaner/cleaning.go
sairameshv: The behavior is same in either of the cases

pkg/cleaner/cleaning.go
```diff
@@ -105,15 +109,8 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
 	// we need to assess whether the file exists already to ensure we don't overwrite existing obfuscated data.
 	// that can happen while obfuscating file names and their paths.
 	// Additionally, the stat check is required because os.O_CREATE will implicitly os.O_TRUNC if a file already exist
```
tjungblu, line 111: last nitpick for today: the comments above now should also go inside the method :)

pkg/cleaner/cleaning.go
tjungblu: this is correct, it just looks weird for the reader as there is no need to supply the type. But it's just my personal preference, so feel free to leave it

pkg/cleaner/cleaning.go
sairameshv: Addressed

pkg/cleaner/cleaning.go
sairameshv: Done

tjungblu (approved):
/lgtm

thanks Ramesh :)

PR #65, opened 2021-09-16
Track how many times patterns are obfuscated

josefkarasek:


pkg/obfuscator/tracker.go
```diff
@@ -15,11 +15,11 @@ type ReplacementTracker interface {
 	Initialize(replacements map[string]string)
 
 	// Report returns a mapping of strings which were replaced.
-	Report() map[string]string
+	Report() map[string]Replacement
 
 	// AddReplacement will add a replacement along with its original string to the report.
 	// If there is an existing value that does not match the given replacement, it will exit with a non-zero status.
-	AddReplacement(original string, replacement string)
+	AddReplacement(original string, replacement Replacement)
```
tjungblu, line 22: leave that at (string, string), if you want to convey a count with multiple replacements then we can add another method that also includes a count, so (string, string, int).

I think it might not be easy for things like the keyword replacer that just replace in bulk - let's see

pkg/obfuscator/tracker.go
```diff
@@ -15,11 +15,11 @@ type ReplacementTracker interface {
 	Initialize(replacements map[string]string)
 
 	// Report returns a mapping of strings which were replaced.
-	Report() map[string]string
+	Report() map[string]Replacement
```
tjungblu, line 18: here's what I would return instead:

```
type ReplacementReport struct {
    replacements []Replacement
}

type Replacement {
   original string
   replaced string
   counter int 
}

// for backwards compatibility in all our tests
func (ReplacementReport) asMap() map[string]string {
}
```

that way we can serialize the list in yaml correctly, can properly sort on the struct level and do diffs in the e2e tests.

tjungblu: in the implementation, you can just keep a second map for the counts and on Report() just collate, similar to the defensive copy we do

pkg/obfuscator/mac_address.go
```diff
@@ -33,12 +33,12 @@ func (m *macAddressObfuscator) Contents(s string) string {
...
 		s = strings.ReplaceAll(s, mac, replacement)
 		// also add the original (non-cleaned) string, this is only used for human review in the final report
-		m.ReplacementTracker.AddReplacement(mac, replacement)
+		m.ReplacementTracker.AddReplacement(mac, Replacement{ReplacedWith: replacement})
 	}
 	return s
 }
 
-func (m *macAddressObfuscator) Report() map[string]string {
+func (m *macAddressObfuscator) Report() map[string]Replacement {
```
tjungblu, line 41: I think we don't even need to define the method here, since the ReplacementTracker is an embedded struct -> candidate for removal

pkg/obfuscator/tracker.go
```diff
@@ -29,42 +39,45 @@ type ReplacementTracker interface {
...
 	if val, ok := s.mapping[original]; ok {
-		if replacement != val {
+		if replacement != val.Replaced {
 			klog.Exitf("'%s' already has a value reported as '%s', tried to report '%s'", original, val, replacement)
 		}
+		val.Total++
 		return
 	}
-	s.mapping[original] = replacement
+	s.mapping[original] = Replacement{Original: original, Replaced: replacement, Total: 0}
```
tjungblu, line 67: that would be Total=1, wouldn't it?

pkg/obfuscator/tracker.go
```diff
@@ -29,42 +39,45 @@ type ReplacementTracker interface {
...
 	if val, ok := s.mapping[key]; ok {
-		return val
+		return val.Replaced
 	}
 	if generator == nil {
 		return ""
 	}
 	r := generator()
-	s.mapping[key] = r
+	s.mapping[key] = Replacement{Original: key, Replaced: r, Total: 0}
```
tjungblu, line 80: same, Total=1?

pkg/obfuscator/keywords.go
```diff
@@ -9,7 +9,7 @@ type keywordsObfuscator struct {
 	replacements map[string]string
 }
 
-func (o *keywordsObfuscator) Report() map[string]string {
+func (o *keywordsObfuscator) Report() ReplacementReport {
```
tjungblu, line 12: just delete those as you go, here it's also embedded

pkg/obfuscator/mac_address.go
```diff
@@ -38,9 +38,9 @@ func (m *macAddressObfuscator) Contents(s string) string {
 	return s
 }
 
-func (m *macAddressObfuscator) Report() map[string]string {
-	return m.ReplacementTracker.Report()
-}
+// func (m *macAddressObfuscator) Reporta() map[string]Replacement {
+// 	return m.Report()
+// }
```
tjungblu, line 43: just delete it :)

pkg/obfuscator/keywords_test.go
```diff
@@ -9,34 +9,34 @@ import (
...
 			name: "partial replacement",
-			replacements: map[string]string{
-				"first-unique":  "first-replacement",
-				"second-unique": "second-replacement",
+			replacements: map[string]Replacement{
+				"first-unique":  Replacement{ReplacedWith: "first-replacement"},
+				"second-unique": Replacement{ReplacedWith: "second-replacement"},
 			},
 			input:          "input with first-unique word",
 			expectedOutput: "input with first-replacement word",
```
tjungblu, line 42: I personally would've left the test and input intact, you just have to change the below Equal to o.Report().AsMap(). Should work the exact same way. No need to change all those tests

pkg/obfuscator/tracker.go
```diff
@@ -29,42 +39,45 @@ type ReplacementTracker interface {
...
+	return ReplacementReport{Replacements: replacements}
 }
 
 func (s *SimpleTracker) AddReplacement(original string, replacement string) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 	if val, ok := s.mapping[original]; ok {
-		if replacement != val {
+		if replacement != val.Replaced {
 			klog.Exitf("'%s' already has a value reported as '%s', tried to report '%s'", original, val, replacement)
```
tjungblu, line 62: val.Replaced also in the log

pkg/obfuscator/tracker_test.go
```diff
@@ -28,7 +28,8 @@ func TestReportLeakingBack(t *testing.T) {
 	tracker := NewSimpleTracker()
 	tracker.AddReplacement("foo", "bar")
 	mapping := tracker.Report()
-	mapping["foo"] = "baz"
+	mapping.asMap()["foo"] = "baz"
```
tjungblu, line 31: should that be possible?

pkg/reporting/reporter.go
```diff
@@ -5,32 +5,34 @@ import (
...
+	Omissions    []string                 `yaml:"omissions,omitempty"`
 }
 
 type Reporter interface {
 	// WriteReport writes the final report into the given path, will create folders if necessary.
 	WriteReport(path string) error
 
 	// CollectOmitterReport will call the Report method on the omitter and collect its omissions.
-	CollectOmitterReport(omitter omitter.ReportingOmitter)
+	CollectOmitterReport(omitter []string)
```
tjungblu, line 23: good catch, please update the comment

pkg/obfuscator/ip.go
```diff
@@ -63,7 +63,7 @@ func (o *ipObfuscator) replace(s string) string {
 				replacement := gen.generateReplacement(cleaned, o.ReplacementTracker)
 				output = strings.ReplaceAll(output, m, replacement)
 				// also add the original (non-cleaned) string, this is only used for human review in the final report
-				o.ReplacementTracker.AddReplacement(m, replacement)
+				o.ReplacementTracker.AddReplacement(cleaned, replacement)
```
josefkarasek, line 66: Since we now care about the uniquenes of keys, 10.131.0.24 != 10-131-0-24

pkg/obfuscator/mac_address.go
```diff
@@ -33,15 +33,11 @@ func (m *macAddressObfuscator) Contents(s string) string {
 		replacement := m.obfsGenerator.generateReplacement(match, m.ReplacementTracker)
 		s = strings.ReplaceAll(s, mac, replacement)
 		// also add the original (non-cleaned) string, this is only used for human review in the final report
-		m.ReplacementTracker.AddReplacement(mac, replacement)
+		m.ReplacementTracker.AddReplacement(match, replacement)
```
josefkarasek, line 36: `0e-a0-e7-92-3a-a3 != 0E:A0:E7:92:3A:A3`

pkg/obfuscator/ip.go
tjungblu: it is that way, so we can later verify in the e2e test that this was replaced. When you have the cleaned version only, then it becomes difficult to reconstruct how it was looking before.

For a human it's also more useful to understand what the actual input was, not a cleaned version. Why do you think we're interested in the uniqueness of the key?

pkg/obfuscator/domain.go
```diff
@@ -48,6 +48,7 @@ func (d *domainObfuscator) replaceDomains(input string) string {
 				replacement = obfuscatedBaseDomain
 			}
 			output = strings.ReplaceAll(output, m[0], replacement)
+			d.ReplacementTracker.AddReplacement(baseDomain, baseDomain, obfuscatedBaseDomain)
```
josefkarasek, line 51: Generators no longer write to the tracker. Each new value must be reported explicitly.

tjungblu: sounds good, do you have the Replacement struct anywhere? I'm not seeing it in the diff

tjungblu, line 51: shouldn't that be:
> d.ReplacementTracker.AddReplacement(baseDomain, m[0], obfuscatedBaseDomain)

?

pkg/obfuscator/ip.go
```diff
@@ -58,12 +58,12 @@ func (o *ipObfuscator) replace(s string) string {
 				continue
 			}
 
-			cleaned := strings.ReplaceAll(m, "-", ".")
+			cleaned := strings.ToUpper(strings.ReplaceAll(m, "-", "."))
```
tjungblu, line 61: not sure what toUpper does with numbers, what's the rationale to add that?

pkg/obfuscator/ip_test.go
```diff
@@ -32,7 +32,7 @@ func TestIPObfuscatorStatic(t *testing.T) {
 			name:   "ipv4 in aws etcd pathing",
 			input:  "must-gather/etcd-ip-10-0-187-218.ec2.internal/some.yaml",
 			output: "must-gather/etcd-ip-xxx.xxx.xxx.xxx.ec2.internal/some.yaml",
-			report: map[string]string{"10-0-187-218": obfuscatedStaticIPv4, "10.0.187.218": obfuscatedStaticIPv4},
+			report: map[string]string{"10-0-187-218": obfuscatedStaticIPv4},
```
tjungblu, line 35: okay so asMap takes the occurrence, right? +1

pkg/obfuscator/keywords.go
```diff
@@ -24,8 +20,8 @@ func (o *keywordsObfuscator) Contents(contents string) string {
 func replace(name string, replacements map[string]string, reporter ReplacementTracker) string {
 	for keyword, replacement := range replacements {
 		if strings.Contains(name, keyword) {
+			reporter.AddReplacementCount(keyword, keyword, replacement, uint(strings.Count(name, keyword)))
 			name = strings.Replace(name, keyword, replacement, -1)
```
tjungblu, line 24: silly question, if we're already doing the counter work, can't we reuse that for the replace call?

so keep the 
> cnt := strings.Count(name, keyword)

and then pass it along the replace:

> name = strings.Replace(name, keyword, replacement, cnt)

pkg/obfuscator/ip.go
josefkarasek: Those are good point, original input will be retained in the report.

pkg/obfuscator/report_types.go
```diff
@@ -0,0 +1,31 @@
...
+	ReplacedWith string       `yaml:"replacedWith,omitempty"`
+	Occurrences  []Occurrence `yaml:"occurrences,omitempty"`
+}
+
+type Occurrence struct {
+	Original string `yaml:"original,omitempty"`
+	Count    uint   `yaml:"count,omitempty"`
+}
+
+func (r *Replacement) Increment(original string, count uint) Replacement {
```
tjungblu, line 18: you should definitely add some tests for this method

tjungblu: and why do you return the Replacement?

pkg/obfuscator/report_types.go
```diff
@@ -0,0 +1,31 @@
...
+	if r.Occurrences == nil {
+		r.Occurrences = []Occurrence{}
+	}
+	for i, o := range r.Occurrences {
```
tjungblu, line 22: this is entirely premature optimization since the whole thing is bottlenecked on regex. why not use a map here?

pkg/obfuscator/target_obfuscator.go
```diff
@@ -21,7 +21,7 @@ func (t *targetObfuscator) Contents(s string) string {
 	return s
 }
 
-func (t *targetObfuscator) Report() map[string]string {
+func (t *targetObfuscator) Report() ReplacementReport {
```
tjungblu, line 24: while you're already touching this piece, can you embed the obfuscator? then we can also kill that reporting method.

pkg/obfuscator/tracker.go
```diff
@@ -15,57 +15,66 @@ type ReplacementTracker interface {
 	Initialize(replacements map[string]string)
 
 	// Report returns a mapping of strings which were replaced.
-	Report() map[string]string
+	Report() ReplacementReport
 
 	// AddReplacement will add a replacement along with its original string to the report.
 	// If there is an existing value that does not match the given replacement, it will exit with a non-zero status.
-	AddReplacement(original string, replacement string)
+	AddReplacement(canonical, original string, replacement string)
```
tjungblu, line 22: please also explain in the comment what canonical is and what difference there is to original

pkg/obfuscator/tracker.go
```diff
@@ -15,57 +15,66 @@ type ReplacementTracker interface {
...
 }
 
 type SimpleTracker struct {
 	lock    sync.RWMutex
-	mapping map[string]string
+	mapping map[string]Replacement
 }
 
-func (s *SimpleTracker) Report() map[string]string {
+var _ ReplacementTracker = (*SimpleTracker)(nil)
```
tjungblu, line 40: what are you trying to accomplish with this?

pkg/obfuscator/tracker.go
```diff
@@ -15,57 +15,66 @@ type ReplacementTracker interface {
...
-func (s *SimpleTracker) AddReplacement(original string, replacement string) {
+func (s *SimpleTracker) AddReplacementCount(canonical, original string, replacement string, count uint) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	if val, ok := s.mapping[original]; ok {
-		if replacement != val {
-			klog.Exitf("'%s' already has a value reported as '%s', tried to report '%s'", original, val, replacement)
-		}
+	if r, ok := s.mapping[canonical]; ok {
+		s.mapping[canonical] = r.Increment(original, count)
```
tjungblu, line 60: okay, now I get that :D why not internally increment the count instead of trying to return the same object?

pkg/obfuscator/tracker.go
```diff
@@ -15,57 +15,66 @@ type ReplacementTracker interface {
...
 		return
 	}
-	s.mapping[original] = replacement
+	new := Replacement{Canonical: canonical, ReplacedWith: replacement}
+	new.Increment(original, count)
```
tjungblu, line 64: I'd add a NewReplacement method that would add the increment internally

pkg/obfuscator/tracker.go
```diff
@@ -15,57 +15,66 @@ type ReplacementTracker interface {
...
+func (s *SimpleTracker) AddReplacementCount(canonical, original string, replacement string, count uint) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
-	if val, ok := s.mapping[original]; ok {
-		if replacement != val {
-			klog.Exitf("'%s' already has a value reported as '%s', tried to report '%s'", original, val, replacement)
-		}
+	if r, ok := s.mapping[canonical]; ok {
+		s.mapping[canonical] = r.Increment(original, count)
 		return
```
tjungblu, line 61: that's a bit nitty, but instead of return you can also put the other clause into an else branch

pkg/obfuscator/tracker.go
```diff
@@ -15,57 +15,66 @@ type ReplacementTracker interface {
...
+	if r, ok := s.mapping[canonical]; ok {
+		return r.ReplacedWith
 	}
 	if generator == nil {
 		return ""
 	}
-	r := generator()
-	s.mapping[key] = r
-	return r
+	return generator()
```
tjungblu, line 77: don't we have to save that in our mapping? Otherwise we will generate a different identifier for the same canonical key.

also be mindful of race conditions, since everything you return here it outside the lock

pkg/obfuscator/tracker.go
```diff
@@ -15,57 +15,66 @@ type ReplacementTracker interface {
...
 }
 
 func (s *SimpleTracker) Initialize(replacements map[string]string) {
```
tjungblu, line 80: I'm not sure whether this is currently used, probably not. But that ideally should take your `ReplacementReport`. I assume that also is touching @sairameshv a bit  - fyi.

pkg/reporting/reporter.go
```diff
@@ -5,33 +5,36 @@ import (
...
+	CollectObfuscatorReport(obfuscatorReport []obfuscator.ReplacementReport)
 }
 
 type SimpleReporter struct {
-	replacements []map[string]string
+	replacements []obfuscator.Replacement
 	omissions    []string
 }
 
+var _ Reporter = (*SimpleReporter)(nil)
```
tjungblu, line 34: what does that do? should we typedef this thing?

pkg/obfuscator/ip.go
josefkarasek: IPv6 addresses

pkg/obfuscator/report_types.go
tjungblu: I mean, you already have a pointer to the replacement, you can just increment the counter directly without returning it.

pkg/obfuscator/ip.go
tjungblu: dang, makes sense. another good reason to split this up. Is there no test case that needs adaptation to this?

pkg/obfuscator/report_types.go
tjungblu: we can turn this into a slice later in the Report method without paying the scan price during runtime

pkg/obfuscator/ip_test.go
josefkarasek: The opposite. AsMap is just the key and replacedWith. it is generated from a structure that for this example is:
```
replacements:
  - canonical: 10.0.187.218
    replacedWith: x-ipv4-0000000001-x
    occurrences:
      - original: 10-0-187-218
        count: 1
```
AsMap returns `.occurences[].original: .replacedWith`

pkg/obfuscator/keywords.go
josefkarasek: Good suggestion, but `strings.Replace()` calls `Count` internally as well. Even if we pass `count` to `Replace`, it will be ignored https://github.com/golang/go/blob/master/src/strings/strings.go#L948

tjungblu: liked the stuff you added on slack, I think we can save ourselves the contains check entirely then.
```
if count := strings.Count(name, keyword); count > 0 { 
  reporter.AddReplacementCount(keyword, keyword, replacement, uint(count))
  name = strings.Replace(name, keyword, replacement, -1)
}
```

pkg/reporting/reporter.go
josefkarasek: Compile time check member functions of SimpleReporter implement functions of interface Reporter

tjungblu: mind blown, why are we not returning the Reporter interface in the New func below?

pkg/obfuscator/domain.go
josefkarasek: I thought that too, but the unit tests were failing when i tried it.
```
fmt.Println(m)
[docs.okd.io docs. okd.io]
...
expected: map[string]string{"okd.io":"domain0000000002", "openshift.com":"domain0000000001"}
actual  : map[string]string{"docs.okd.io":"domain0000000002", "openshift.com":"domain0000000001"}

```

josefkarasek: Might it be that the unit tests have different expectations?

pkg/obfuscator/report_types.go
josefkarasek: :+1: on tests

I'll review how Replacement is used

pkg/obfuscator/domain.go
tjungblu: the map representation now changed a bit, the report will show the right thing. and there is also the side-effect of the generator changes below.

I reckon the tests here then should use the Report directly instead of the map.

pkg/obfuscator/report_types.go
tjungblu: also below in the reporting is nothing else like a good old std::multimap. I'm sure there must be a good library for us to reuse?

pkg/obfuscator/keywords.go
josefkarasek: I wouldn't do that. `Index()` is potentially very CPU heavy operation. `Contains()` calls `Index()` once, `Count()` calls `Index()` in a for loop.

Now we call `Count()` only when `Contains() == true`, thus works as a flood gate before we call `Count()`

pkg/obfuscator/tracker.go
josefkarasek: I modified `AddReplacement` so that it takes the canonical key (`cleaned` version for IP and MAC) along with the original key

```
AddReplacement(canonical, original string, replacement string)
```

pkg/obfuscator/report_types.go
josefkarasek: 1. I'm not concerned about the performance here: Eg for MAC addresses we can have up to 4 elements in the slice (dots, hyphens, upper/lower case)
2. I can change it to a map, but I don't think we'll benefit from it at all.

pkg/obfuscator/tracker.go
tjungblu: the reason it was there, is because there can be a race condition where two goroutines find an empty mapping for the canonical string, generator then creates two representations for the same string.

having this here under a single locking was guaranteed that this could never happen. You also removed that exitf that was basically covering that case if it would happen. Put that exitf back and try it out on a couple of MGs, it should happen fairly frequently.

pkg/obfuscator/report_types.go
tjungblu: you'd be surprised, but I found a must-gather with more than a million IPs:
https://github.com/openshift/must-gather-clean/pull/50

pkg/reporting/reporter.go
josefkarasek: Good point. I'll change that.

pkg/obfuscator/tracker.go
josefkarasek: added

pkg/obfuscator/tracker.go
josefkarasek: added

pkg/obfuscator/tracker.go
josefkarasek: We'll need the map to be a map of pointers for this to work.
```
type SimpleTracker struct {
	lock    sync.RWMutex
	mapping map[string]*Replacement
}
```

pkg/obfuscator/domain.go
josefkarasek: I changed the unit tests to use the actual reports

pkg/obfuscator/ip.go
josefkarasek: I added 2 new test cases

pkg/obfuscator/tracker.go
```diff
@@ -85,12 +87,22 @@ func (s *SimpleTracker) GenerateIfAbsent(canonical string, original string, coun
 	return g
 }
 
-func (s *SimpleTracker) Initialize(replacements map[string]string) {
+func (s *SimpleTracker) Initialize(report ReplacementReport) {
+	close(onlyOneInit) // panics when called twice
```
tjungblu, line 91: nifty!

PR #67, opened 2021-09-21
Re-use of Obfuscation reports

sairameshv:
Contains the code changes to re-use the previously generated reports.

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -34,12 +34,12 @@ func RunPipe(configPath string, stdin io.Reader, stdout io.Writer) error {
 			return fmt.Errorf("failed to create obfuscators via config at %s: %w", configPath, err)
 		}
 	} else {
-		ipObfuscator, err := obfuscator.NewIPObfuscator(schema.ObfuscateReplacementTypeConsistent)
+		ipObfuscator, err := obfuscator.NewIPObfuscator(schema.ObfuscateReplacementTypeConsistent, map[string]string{})
 		if err != nil {
 			return fmt.Errorf("failed to create IP obfuscator: %w", err)
 		}
 
-		macObfuscator, err := obfuscator.NewMacAddressObfuscator(schema.ObfuscateReplacementTypeConsistent)
+		macObfuscator, err := obfuscator.NewMacAddressObfuscator(schema.ObfuscateReplacementTypeConsistent, map[string]string{})
 		if err != nil {
 			return fmt.Errorf("failed to create MAC obfuscator: %w", err)
 		}
@@ -151,24 +151,24 @@ func createObfuscatorsFromConfig(config *schema.SchemaJson) (*obfuscator.MultiOb
 		)
 		switch o.Type {
 		case schema.ObfuscateTypeKeywords:
-			k = obfuscator.NewKeywordsObfuscator(o.Replacement)
+			k = obfuscator.NewKeywordsObfuscator(o.Replacement, o.Report)
 		case schema.ObfuscateTypeMAC:
-			k, err = obfuscator.NewMacAddressObfuscator(o.ReplacementType)
+			k, err = obfuscator.NewMacAddressObfuscator(o.ReplacementType, o.Report)
 			if err != nil {
 				return nil, err
 			}
 		case schema.ObfuscateTypeRegex:
-			k, err = obfuscator.NewRegexObfuscator(*o.Regex)
+			k, err = obfuscator.NewRegexObfuscator(*o.Regex, o.Report)
 			if err != nil {
 				return nil, err
 			}
 		case schema.ObfuscateTypeDomain:
-			k, err = obfuscator.NewDomainObfuscator(o.DomainNames, o.ReplacementType)
+			k, err = obfuscator.NewDomainObfuscator(o.DomainNames, o.ReplacementType, o.Report)
 			if err != nil {
 				return nil, err
 			}
 		case schema.ObfuscateTypeIP:
-			k, err = obfuscator.NewIPObfuscator(o.ReplacementType)
+			k, err = obfuscator.NewIPObfuscator(o.ReplacementType, o.Report)
 			if err != nil {
 				return nil, err
 			}
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -3,11 +3,12 @@ package cli
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/openshift/must-gather-clean/pkg/cleaner"
-	"github.com/openshift/must-gather-clean/pkg/reporting"
+	"gopkg.in/yaml.v2"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/must-gather-clean/pkg/obfuscator"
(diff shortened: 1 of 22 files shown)


sairameshv:
@tjungblu , requesting for your review when you find time. I'm unable to add you to the reviewer's list.

pkg/cli/cli.go
```diff
@@ -3,11 +3,12 @@ package cli
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/openshift/must-gather-clean/pkg/cleaner"
-	"github.com/openshift/must-gather-clean/pkg/reporting"
+	"gopkg.in/yaml.v2"
```
tjungblu, line 11: I think we're using v3 everywhere

pkg/cli/cli.go
```diff
@@ -132,40 +145,36 @@ func createOmittersFromConfig(config *schema.SchemaJson) (omitter.ReportingOmitt
...
-			k, err := obfuscator.NewIPObfuscator(o.ReplacementType)
+			k, err = obfuscator.NewIPObfuscator(o.ReplacementType, o.Report)
 			if err != nil {
 				return nil, err
 			}
-			k = obfuscator.NewTargetObfuscator(o.Target, k)
-			obfuscators = append(obfuscators, k)
 		}
+		k = obfuscator.NewTargetObfuscator(o.Target, k)
+		obfuscators = append(obfuscators, k)
```
tjungblu, line 177: this is a great refactoring, thanks

pkg/obfuscator/obfuscator.go
```diff
@@ -6,6 +6,8 @@ type Obfuscator interface {
 	Path(string) string
 	// Contents takes string as input and return the obfuscated string
 	Contents(string) string
+	// Type returns an obfuscator type which would come into help during the reporting.
+	Type() string
```
tjungblu, line 10: did you actually make use of this?

pkg/cli/cli.go
```diff
@@ -132,40 +145,36 @@ func createOmittersFromConfig(config *schema.SchemaJson) (omitter.ReportingOmitt
...
+		var (
+			k   obfuscator.ReportingObfuscator
+			err error
+		)
 		switch o.Type {
 		case schema.ObfuscateTypeKeywords:
-			k := obfuscator.NewKeywordsObfuscator(o.Replacement)
-			k = obfuscator.NewTargetObfuscator(o.Target, k)
-			obfuscators = append(obfuscators, k)
+			k = obfuscator.NewKeywordsObfuscator(o.Replacement, o.Report)
```
tjungblu, line 154: as mentioned on slack, just create an already initialized tracker instead of passing the raw map around. You already did a great job here to refactor the loop, so you can just init at the top and pass down.

pkg/cli/cli.go
```diff
@@ -90,14 +91,26 @@ func Run(configPath string, inputPath string, outputPath string, deleteOutputFol
 
 	traversal.NewParallelFileWalker(inputPath, workerCount, workerFactory).Traverse()
 
-	reporter := reporting.NewSimpleReporter()
-	reporter.CollectOmitterReport(mro.Report())
-	reporter.CollectObfuscatorReport(mo.ReportPerObfuscator())
-	reporterErr := reporter.WriteReport(filepath.Join(reportingFolder, reportFileName))
-	if reporterErr != nil {
-		return reporterErr
+	mo.UpdateReportPerObfuscator(config)
```
tjungblu, line 94: where's the omission?

pkg/cli/cli.go
```diff
@@ -90,14 +91,26 @@ func Run(configPath string, inputPath string, outputPath string, deleteOutputFol
...
+
+	/*
+		reporter := reporting.NewSimpleReporter()
+		reporter.CollectOmitterReport(mro.Report())
+		reporter.CollectObfuscatorReport(mo.ReportPerObfuscator())
+		reporterErr := reporter.WriteReport(filepath.Join(reportingFolder, reportFileName))
+		if reporterErr != nil {
+			return reporterErr
+		}
+	*/
```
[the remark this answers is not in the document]
tjungblu: I love that you remove code, I think we still need it (very least for the omissions) and the business logic in `UpdateReportPerObfuscator` should go into this reporting package. I'd extend this by adding a config to the reporter and then it can hash out what to merge with which result.

pkg/obfuscator/multi_obfuscator.go
```diff
@@ -41,6 +49,19 @@ func (m *MultiObfuscator) ReportPerObfuscator() []map[string]string {
 	return multiReport
 }
 
+func (m *MultiObfuscator) UpdateReportPerObfuscator(config *schema.SchemaJson) {
```
tjungblu, line 52: that should go into the reporting package

pkg/obfuscator/obfuscator.go
tjungblu: haven't found a reference, so let's remove?

pkg/obfuscator/tracker.go
```diff
@@ -74,8 +74,11 @@ func (s *SimpleTracker) Initialize(replacements map[string]string) {
 	if len(s.mapping) > 0 {
 		klog.Exitf("tracker was initialized more than once or after some replacements were already added.")
 	}
-	for k, v := range replacements {
-		s.mapping[k] = v
+
+	if len(replacements) > 0 {
```
tjungblu, line 78: why the length check? the for loop would just not execute when the map is empty

pkg/schema/schema.go
```diff
@@ -2,36 +2,38 @@
```
tjungblu, line 2: didn't you have to change the json file?

pkg/schema/schema.go
```diff
@@ -40,7 +42,10 @@ type Obfuscate struct {
 	// static replacement where a detected mac address will be replaced by 'x'. Regex
 	// should be used with the 'regex' property that will define the regex, here the
 	// replacement also will be static by 'x'-ing out the matched string.
-	Type ObfuscateType `json:"type"`
+	Type ObfuscateType `json:"type" yaml:"type"`
+
+	// Report represents the replacement mapping of the obfuscated data
+	Report map[string]string `json:"report" yaml:"report"`
```
tjungblu, line 48: yeah, that doesn't work :D This is generated code from the schema defined in json.

pkg/schema/schema.go
tjungblu, line 2: can you please also update the README with the new possibility to "pre-set" those values?

pkg/schema/schema.go
tjungblu: we shouldn't call it `Report` either, maybe we can reuse the `ObfuscateReplacement` map for this? We can adjust the reporting to populate those fields.

pkg/cli/cli.go
sairameshv: Okay

pkg/obfuscator/obfuscator.go
sairameshv: This is getting used in `"UpdateReportPerObfuscator"` function where the report for each obfuscate type would be accordingly appended

pkg/cli/cli.go
sairameshv: Sure

pkg/cli/cli.go
sairameshv: The code still needs to be added

pkg/schema/schema.go
sairameshv: Yes, I have to change

pkg/schema/schema.go
sairameshv: I would like to discuss on concluding the schema.json file based on the recent merges and accordingly generate this file.

sairameshv:
Closing this PR to raise a new one with the recently merged changes :)

PR #68, opened 2021-09-22
support for re-use of the obfuscated reports

sairameshv:
These changes help in re-using the reports generated by the previous obfuscations.

### DIFF
--- a/pkg/obfuscator/multi_obfuscator_test.go
+++ b/pkg/obfuscator/multi_obfuscator_test.go
@@ -60,7 +60,7 @@ func TestMultiObfuscationPaths(t *testing.T) {
 func TestMultiObfuscationReport(t *testing.T) {
 	mo := NewMultiObfuscator(
 		[]ReportingObfuscator{
-			&splitObfuscator{tracker: NewSimpleTracker()},
+			&splitObfuscator{tracker: NewSimpleTracker(map[string]string{})},
 		})
 
 	contents := mo.Contents("this must be split once")
@@ -81,9 +81,9 @@ func TestMultiObfuscationReportShouldOverride(t *testing.T) {
 func TestMultiObfuscationReportMulti(t *testing.T) {
 	mo := NewMultiObfuscator(
 		[]ReportingObfuscator{
-			&splitObfuscator{tracker: NewSimpleTracker()},
-			&splitObfuscator{tracker: NewSimpleTracker()},
-			&splitObfuscator{tracker: NewSimpleTracker()},
+			&splitObfuscator{tracker: NewSimpleTracker(map[string]string{})},
+			&splitObfuscator{tracker: NewSimpleTracker(map[string]string{})},
+			&splitObfuscator{tracker: NewSimpleTracker(map[string]string{})},
 		})
 
 	contents := mo.Contents("this must be split thrice")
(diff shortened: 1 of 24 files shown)


pkg/obfuscator/multi_obfuscator_test.go
```diff
@@ -60,7 +60,7 @@ func TestMultiObfuscationPaths(t *testing.T) {
 func TestMultiObfuscationReport(t *testing.T) {
 	mo := NewMultiObfuscator(
 		[]ReportingObfuscator{
-			&splitObfuscator{tracker: NewSimpleTracker()},
+			&splitObfuscator{tracker: NewSimpleTracker(map[string]string{})},
```
tjungblu, line 63: now it's a bit too late, but you could've added another New* method instead of filling all the defaults here.

`NewSimpleTracker()` could just use an empty map underneath and `NewSimpleTrackerMap(map...)` would be a new method based off of a map. Less code to change :)

pkg/obfuscator/tracker.go
```diff
@@ -106,6 +106,15 @@ func (s *SimpleTracker) Initialize(report ReplacementReport) {
 	}
 }
 
-func NewSimpleTracker() ReplacementTracker {
-	return &SimpleTracker{mapping: map[string]*Replacement{}}
+func NewSimpleTracker(existingReplacements map[string]string) ReplacementTracker {
+	var tracker = &SimpleTracker{mapping: map[string]*Replacement{}}
+	tracker.lock.Lock()
```
tjungblu, line 111: if you reverse the order of the statements, you can avoid the locking:

```
var m map[string]*Replacement{}        
for key, value := range existingReplacements {
   m[key] = NewReplacement(key, key, value, 0)
}
return &SimpleTracker{mapping: m}
```

tjungblu: also to mention: there is no way any other goroutine could've changed your tracker here, so that lock is also not necessary in the first place

pkg/reporting/reporter_test.go
```diff
@@ -7,13 +7,15 @@ import (
...
 	"github.com/openshift/must-gather-clean/pkg/obfuscator"
+	"github.com/openshift/must-gather-clean/pkg/schema"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
 )
 
 func TestReportingHappyPath(t *testing.T) {
-	r := NewSimpleReporter()
+	config := new(schema.SchemaJson)
```
tjungblu, line 17: can you please fill this with some sensible information and update the asserts below?

tjungblu (changes requested):
rest looks great, thanks Ramesh :)

pkg/obfuscator/multi_obfuscator_test.go
sairameshv: Done

pkg/reporting/reporter.go
```diff
@@ -56,10 +59,24 @@ func (s *SimpleReporter) WriteReport(path string) error {
 	if err != nil {
 		return fmt.Errorf("failed to open report file %s: %w", path, err)
 	}
+
+	for i, obfs := range s.config.Config.Obfuscate {
```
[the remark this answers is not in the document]
sairameshv: Addressed

tjungblu (approved):
/lgtm

thanks @sairameshv!

Issue #78, opened 2021-10-06
broken symlinks are not managed/handled


sferich888:
In theory; due to the way this tool is written; **any archive can be passed into the tool**. As such I wanted to see if this could be used as a replacement (or alternative options for `sos-clean`), mainly because of the configurabilty of this tool, vs that `sos-clean`. 

In theory this should work, however due to the way sosreport generates its 'archive' you have broken symlink file references to contend with.

Example: 
```
$ file sosreport-ip-10-0-xxx-xxx-YYYY-MM-DD-FOO/etc/grub.d/15_ostree
sosreport-ip-10-0-xxx-xxx-YYYY-MM-DD-FOO/etc/grub.d/15_ostree: broken symbolic link to ../../usr/libexec/libostree/grub2-15_ostree
```

Due to this; when you try to run `must-gather-clean` against a sosreport; you're presented with a fatal error:  

```
$ ./must-gather-clean -c openshift_default.yaml -i sosreport-ip-10-0-xxx-xxx-YYYY-MM-DD-FOO -o sosreport_cleaned
F1006 HH:MM:SS.MS 1062736 traversal.go:43] failed to process etc/grub.d/15_ostree due to open sosreport-ip-10-0-xxx-xxx-YYYY-MM-DD-FOO/etc/grub.d/15_ostree: no such file or directory
```
Error: https://github.com/openshift/must-gather-clean/blob/main/pkg/traversal/traversal.go#L43 (which is sorta expected). 

**What did you expect to see?**
A clear and concise description of what you expected to happen (or insert a code snippet).

**What did you see instead? Under which circumstances?**
A clear and concise description of what you expected to happen (or insert a code snippet).

**Environment**
* must-gather-clean version:  `Version: pkg.Version{Version:"v0.0.1-0-gd02283f", GitCommit:"d02283f", BuildDate:"2021-10-01T09:44:36Z", GoOs:"linux", GoArch:"amd64"}`

**Possible Solution**
To handle these files; we may want to default to 'omitting' them (or working an option into the omit config for this), or simply copy over the file name/reference to the output archive (as they are blank/broken) and shouldn't need further processing. 

Either way adustments in https://github.com/openshift/must-gather-clean/blob/main/pkg/cleaner/cleaning.go#L60-L93 are probably needed.

sferich888:
This should be resolved by https://github.com/openshift/must-gather-clean/pull/80

PR #79, opened 2021-10-11
ignore symbolic links during traversal

tjungblu:
fix for #78 where the symbolic link handling was assuming all links were relative and it broke on absolute paths. 
This PR ignores them and prints a log statement when it is encountered.

### DIFF
--- a/pkg/traversal/traversal_test.go
+++ b/pkg/traversal/traversal_test.go
@@ -2,6 +2,7 @@ package traversal
 
 import (
 	"path/filepath"
+	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


pkg/traversal/traversal_test.go
```diff
@@ -45,6 +45,9 @@ func TestFileWalker(t *testing.T) {
 
 			walker.Traverse()
 
+			sort.Strings(tc.expectedResult)
+			sort.Strings(queueProc.paths)
+
 			assert.Equal(t, tc.expectedResult, queueProc.paths)
```
josefkarasek, line 51: [testify.ElementsMatch](https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch) ignores order

PR #86, opened 2023-05-25
Replace default regex lib with pcre regex lib

bachoudev:
I have a context where i need to match all string of 6 or 8 char (with number) in a yaml file. 
This is a real problem today because he match only on the value in yaml and not key - value. 

There is a solution where i can use a lookahead or behind to restrict the match of a regex but the default go library can't handle this kind of syntax since RE2 not supported it. 

So we made for this usage a fork with pcre syntax support. We can now use lookahead or behind to handle this kind of context.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/atombender/go-jsonschema v0.0.0-00010101000000-000000000000
+	github.com/gijsbers/go-pcre v0.0.0-20161214203829-a84f3096ab3c // indirect
 	github.com/openshift/build-machinery-go v0.0.0-20210712174854-1bb7fd1518d3
 	github.com/spf13/cobra v1.1.3
 	github.com/stretchr/testify v1.7.0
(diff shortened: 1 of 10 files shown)


lmzuccarelli (commented):
The internal prow ci test is failing due to go.mod content being incorrect.

Could you please re submit but do a `go mod tidy` first

bachoudev:
@lmzuccarelli done & pushed

lmzuccarelli:
@bachoudev - thanks for contributing

lmzuccarelli (approved):

tjungblu:
@bachoudev and @lmzuccarelli this library does not compile for MAC (amd64 + ARM) and ARM on all platforms, what was the rationale to use this? That's a breaking change and I can't create a 0.0.2 release with this.

bachoudev:
@tjungblu the main objectif was to benefit from pcre to use look arround and other regex features wich are not supported by RE2 syntaxe in go. 
This change was needed for our usage but maybe we need to fork this on our hand and revert this change like your PR proposal #91 for now.
And for the future i can check if it's possible to edit the source lib to make it work with other arch.

PR #100, opened 2025-09-16
Obfuscate clusterid

janboll:
Spike, needs some testing

Adding an obfuscator for cluster ids

### DIFF
--- a/examples/openshift_default.yaml
+++ b/examples/openshift_default.yaml
@@ -15,6 +15,9 @@ config:
     - type: AzureResources
       replacementType: Consistent
       target: All
+    - type: ClusterId
+      replacementType: Consistent
+      target: All
   omit:
     - type: Kubernetes
       kubernetesResource:
--- a/pkg/obfuscator/cluster_id.go
+++ b/pkg/obfuscator/cluster_id.go
@@ -1,0 +1,58 @@
+package obfuscator
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/openshift/must-gather-clean/pkg/schema"
+)
+
+var (
+	// From OCM codebase, pkg/models/uid.go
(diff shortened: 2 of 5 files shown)


pkg/obfuscator/cluster_id.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/openshift/must-gather-clean/pkg/schema"
+)
+
+var (
+	// From OCM codebase, pkg/models/uid.go
+	ocmClusterIdPattern            = "[0123456789abcdefghijklmnopqrstuv]{32}"
```
janboll, line 13: So, this might produce false hits, i.e. when there is some random text value like a sha: `2b4a2379dafaef96fce39c29b7ab2b6424x`
That string will be redacted as well. 

Should I keep it to have some false hits or should I rather be more strict on matching?

examples/openshift_default.yaml
```diff
@@ -15,6 +15,9 @@ config:
     - type: AzureResources
```
deads2k, line 15: Can we extend the AzureResources to avoid producing lots config elements?

pkg/obfuscator/cluster_id.go
```diff
@@ -0,0 +1,58 @@
...
+	clusterIDReplacement           = "obfuscatedclusterid%013d"
```
deads2k, line 14: the x-foo-x pattern is preferred

examples/openshift_default.yaml
deads2k: It's still impractical for us to add add more and more obfuscators.  Let's try combining now and perhaps we rename?

janboll: I don't see how this is impractical, but I do not want to die on this hill. So I refactored the clusterid part into the Azure Resource Obfuscator. Can you please take another look?

PR #109, opened 2026-02-12
make the azure replacements adjectiveNoun pairs for easier consumption

deads2k:
Found this while actually using the redacted must-gather.  It's useful to have names that can be pronounced and more easily identified when reading a log.  This uses the docker style of names.

### DIFF
--- a/pkg/obfuscator/azure_resources.go
+++ b/pkg/obfuscator/azure_resources.go
@@ -7,7 +7,6 @@ import (
 	"sort"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/openshift/must-gather-clean/pkg/schema"
 	"k8s.io/klog/v2"
@@ -138,9 +137,11 @@ func (o *azureResourceObfuscator) replace(s string) string {
 	return patternReplacedString
 }
 
-func NewAzureResourceObfuscator(replacementType schema.ObfuscateReplacementType, tracker ReplacementTracker, seed int64) (ReportingObfuscator, error) {
-	if seed == 0 {
-		seed = time.Now().UnixNano()
+func NewAzureResourceObfuscator(replacementType schema.ObfuscateReplacementType, tracker ReplacementTracker, desiredSeed *int) (ReportingObfuscator, error) {
+	var randSource RandomSource
+	randSource = cryptoRandSource{}
+	if desiredSeed != nil {
+		randSource = rand.New(rand.NewSource(int64(*desiredSeed)))
 	}
 
 	if replacementType != schema.ObfuscateReplacementTypeStatic && replacementType != schema.ObfuscateReplacementTypeConsistent {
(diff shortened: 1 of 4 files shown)


pkg/obfuscator/azure_resources.go
```diff
@@ -144,99 +138,84 @@ func (o *azureResourceObfuscator) replace(s string) string {
...
-		must(newGenerator(azureSubscriptionTemplate, staticAzureSubscriptionReplacement, maximumSupportedObfuscationsAzure, replacementType)),
-		func(original string, matches []string, replacer *partialRegexReplacer) string {
-			if len(matches) < 2 {
-				return original
-			}
+				providerName := matches[1]
+				resourceType := matches[2]
+				resourceName := matches[3]
+				subresourceType := matches[4]
+				subresourceNameReplacement := replacer.generateReplacement(matches[5], matches[5], 1, tracker)
```
janboll: For subscriptions, what if you simply generate a new uuid instead? It would be following same regex/reading patterns.

pkg/obfuscator/artifacts/adjectives.txt
```diff
@@ -0,0 +1,449 @@
+able
```
tjungblu: are there any licensing concerns? not sure where you got that list from

pkg/obfuscator/petname_generator.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+type PetNameGenerator struct {
+	separator string
+
+	adverbs    []string
+	adjectives []string
+	names      []string
+
+	rand *rand.Rand
```
tjungblu: Given this is used by some rather security minded customers, crypto/rand would probably the best choice here.

eg `rand.Int(rand.Reader, big.NewInt(100))` 


I wanted to initially call out that this is not thread-safe either, but the tracker/generator function is under a lock, so nothing to worry about

pkg/obfuscator/azure_resources.go
```diff
@@ -144,99 +138,84 @@ func (o *azureResourceObfuscator) replace(s string) string {
 	return patternReplacedString
 }
 
-func must[T any](v T, err error) T {
-	if err != nil {
-		panic(err)
+func NewAzureResourceObfuscator(replacementType schema.ObfuscateReplacementType, tracker ReplacementTracker, seed int64) (ReportingObfuscator, error) {
+	if seed == 0 {
```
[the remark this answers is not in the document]
deads2k: Ok

pkg/obfuscator/azure_resources.go
deads2k: > For subscriptions, what if you simply generate a new uuid instead? It would be following same regex/reading patterns.

I suspect that will make people less certain they got redacted.

PR #114, opened 2026-04-21
fix: prevent obfuscation false positives on common words and substrings

trevorwilliams2025:
## Summary      
                                                                                                                                                                                
  - Azure resource names like "service", "network", or "proxy" discovered from ARM                                                                                              
    paths were being blindly replaced in all free text via strings.ReplaceAll, corrupting                                                                                       
    unrelated strings (e.g. "containerd.service" → "containerd.obfuscated-resource-name")                                                                                       
  - Keywords obfuscator had the same substring problem — a keyword like "dns" would match                                                                                       
    inside "coredns"                                                                                                                                                            
                                                                                                                                                                                
  ## Fix                                                                                                                                                                        
                                                                                                                                                                                
  - **Keywords obfuscator**: switched from strings.ReplaceAll to \b-bounded regex matching,                                                                                     
    so replacements respect word boundaries (underscores and digits are word characters;
    dots and hyphens are boundaries)                                                                                                                                            
  - **Azure resource obfuscator**: added two guards for free-text replacement:                                                                                                  
    1. `replaceNotInsideWord` — skips matches embedded inside larger alphabetic tokens                                                                                          
       (e.g. "Proxy1" inside "MyProxy1Handler")                                                                                                                                 
    2. `isGenericWord` — skips single-case alphabetic strings ("service", "GPU") that are                                                                                       
       too common to safely replace in free text                                                                                                                                
    3. Short canonicals (< 5 chars) are skipped entirely ("0", "vm", "rg")                                                                                                      
                                                                                                                                                                                
  ## Test plan                                                                                                                                                                  
                                                                                                                                                                                
  - [ ] New test: "service" discovered from ARM path does not corrupt "containerd.service"                                                                                      
  - [ ] New test: keyword "dns" does not match inside "coredns"
  - [ ] Edge case tests for: mixed-case identifiers, digit suffixes, URL-encoded strings,                                                                                       
        hyphenated names, multiple occurrences, embedded-in-token protection                                                                                                    
  - [ ] Word boundary semantics verified for: dots, hyphens, underscores, colons, equals, digits                                                                                
  - [ ] `go build ./...` and `go test ./pkg/obfuscator/...` pass

### DIFF
--- a/pkg/obfuscator/azure_resources.go
+++ b/pkg/obfuscator/azure_resources.go
@@ -9,7 +9,6 @@ import (
 	"sync"
 
 	"github.com/openshift/must-gather-clean/pkg/schema"
-	"k8s.io/klog/v2"
 	"k8s.io/utils/set"
 )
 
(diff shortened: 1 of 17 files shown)


pkg/obfuscator/azure_resources.go
```diff
@@ -20,6 +19,66 @@ const (
 	staticAzureSubresourceNameReplacement = "obfuscated-subresource-name"
 )
 
+func isLetter(b byte) bool {
+	return (b >= 'a' && b <= 'z') || (b >= 'A' && b <= 'Z')
```
janboll, line 23: what about cases like `service-foo`, that would be replaced to `obfuscated-foo`?

pkg/obfuscator/azure_resources.go
```diff
@@ -20,6 +19,66 @@ const (
...
+}
+
+// replaceNotInsideWord is like strings.ReplaceAll but skips matches embedded
+// inside a larger alphabetic token (e.g. "Proxy1" inside "MyProxy1Handler").
+func replaceNotInsideWord(s, old, repl string) (uint, string) {
```
janboll, line 28: this needs more documentation or individual unit test to better understand how it's functioning

pkg/obfuscator/azure_resources.go
trevorwilliams2025: Replying to @janboll's comment on `isLetter` (line 23):

> what about cases like `service-foo`, that would be replaced to `obfuscated-foo`?

This is handled by two layers of protection:

1. **`genericSkipWords`** — `service` is in the skip list ([protected.go:38](https://github.com/openshift/must-gather-clean/blob/d799b6c/pkg/obfuscator/protected.go#L38)), so it's never replaced in free text even if discovered as an ARM resource name.

2. **End-to-end test** — `TestAzureResourceObfuscatorContents` has a case called "common word resource name does not corrupt hyphenated compounds" ([azure_resources_test.go:733-741](https://github.com/openshift/must-gather-clean/blob/d799b6c/pkg/obfuscator/azure_resources_test.go#L733-L741)) that proves exactly this:
   - Input: ARM path discovers `service` as a resource name, then free text contains `service-foo is ready`
   - Output: `service-foo is ready` — unchanged

---

Replying to @janboll's comment on `replaceNotInsideWord` (line 28):

> this needs more documentation or individual unit test to better understand how it's functioning

The function has been renamed to `replaceStandalone` and now has:
- Detailed doc comment explaining boundary rules with examples ([azure_resources.go:27-35](https://github.com/openshift/must-gather-clean/blob/d799b6c/pkg/obfuscator/azure_resources.go#L27-L35))
- Dedicated `TestReplaceStandalone` with 12 cases covering spaces, equals signs, string boundaries, letter neighbors, hyphens, digits, and edge cases ([azure_resources_test.go:271-399](https://github.com/openshift/must-gather-clean/blob/d799b6c/pkg/obfuscator/azure_resources_test.go#L271-L399))

PR #118, opened 2026-05-27
Add tar-aware processing for .tar.gz and .tgz archives.

mrobson:
Previously, tar.gz/.tgz files were decompressed as plain gzip and the raw tar binary stream was treated as text for obfuscation, which was incorrect and cause the cleaned file to become corrupted. Now tar entries are iterated individually and each file entry is obfuscated separately.

Added some tests as well.

Example:

> tar zxvf ovnk_database_store.tar.gz
> x ovnk_database_store/
> x ovnk_database_store/ovnkube-node-2s8ft_sbdb
> tar: Damaged tar archive
> tar: Retrying...
> tar: Damaged tar archive
> tar: Retrying...

### DIFF
--- a/pkg/cleaner/cleaning.go
+++ b/pkg/cleaner/cleaning.go
@@ -142,23 +142,8 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
 		}
 	}
 
-	// tar archives need entry-level processing instead of treating the whole binary stream as text
-	if strings.HasSuffix(readPath, ".tar.gz") || strings.HasSuffix(readPath, ".tgz") {
-		err = c.obfuscateTarGz(inputOsFile, outputOsFile, readPath, reportOnly)
-		if err != nil {
-			return fmt.Errorf("failed to obfuscate tar archive '%s': %w", readPath, err)
-		}
-		if err = inputOsFile.Close(); err != nil {
-			return fmt.Errorf("failed to close input file '%s': %w", readPath, err)
-		}
-		if err = outputOsFile.Close(); err != nil {
-			return fmt.Errorf("failed to close output file '%s': %w", writePath, err)
-		}
-		return nil
-	}
-
-	// must-gathers can include gunzipped log files nowadays, handling this special case here once
-	if strings.HasSuffix(readPath, ".gz") {
+	// must-gathers can include gzipped files, handling decompression/recompression here
+	if strings.HasSuffix(readPath, ".gz") || strings.HasSuffix(readPath, ".tgz") {
 		inputOsFile, err = gzip.NewReader(inputOsFile)
 		if err != nil {
 			return fmt.Errorf("failed to create a gzip reader when opening '%s': %w", readPath, err)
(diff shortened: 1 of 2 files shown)


pkg/cleaner/cleaning.go
```diff
@@ -189,6 +206,73 @@ func (c *FileContentObfuscator) createNonConflictingFileUnderLock(outputFilePath
...
+	}
+	defer gzReader.Close()
+
+	tarReader := tar.NewReader(gzReader)
+
+	var tarWriter *tar.Writer
+	var gzWriter *gzip.Writer
+	if !reportOnly {
+		gzWriter = gzip.NewWriter(output)
+		tarWriter = tar.NewWriter(gzWriter)
```
tjungblu: in this case you can use defer, what I use nowadays is a joined error with a named return type.


```suggestion
		gzWriter = gzip.NewWriter(output)
		tarWriter = tar.NewWriter(gzWriter)
		
		defer func() {
		   err = errors.Join(err, tarWriter.Close(), gzWriter.Close())
		}
```

That way you don't actually leak anything (not much of a problem, it's a short lived CLI) and retain all errors.

We didn't have joined errors back then and this individual close pattern is spread everywhere, but it may be worth using them from now on :)

tjungblu: You can probably do the same with the gz reader.

pkg/cleaner/cleaning.go
```diff
@@ -189,6 +206,73 @@ func (c *FileContentObfuscator) createNonConflictingFileUnderLock(outputFilePath
...
+			}
+			continue
+		}
+
+		if reportOnly {
+			if err := c.ObfuscateReader(tarReader, io.Discard); err != nil {
+				return fmt.Errorf("failed to obfuscate tar entry '%s' in '%s': %w", header.Name, readPath, err)
+			}
+			continue
+		}
```
tjungblu: how much of a performance impact does it have to just skip this short-circuit check against the `reportOnly` flag? 

The reason I'm asking is because we have this no-op discard writer:

``` 
	if reportOnly {
		outputOsFile = nopCloser{io.Discard}
```  

so all the logic later does not need to worry about what reporting means. In the normal case it's not a big deal, but I assume the tar and gz writing would be more expensive.

pkg/cleaner/cleaning.go
```diff
@@ -140,8 +142,23 @@ func (c *FileContentObfuscator) ObfuscateFile(inputFile string, outputFile strin
...
+			return fmt.Errorf("failed to obfuscate tar archive '%s': %w", readPath, err)
+		}
+		if err = inputOsFile.Close(); err != nil {
+			return fmt.Errorf("failed to close input file '%s': %w", readPath, err)
+		}
+		if err = outputOsFile.Close(); err != nil {
+			return fmt.Errorf("failed to close output file '%s': %w", writePath, err)
+		}
+		return nil
+	}
```
tjungblu: interesting choice, I would've layered this after the gzip reader below. That way you can just get the zipping logic shared. Imagine we would have .tar.bz2 files (urgh), then we can just add another branch reader/writer setup there and it would be totally opaque for the tar reading.

pkg/cleaner/cleaning.go
mrobson: Not much ya... but no longer relevant with the changes.

pkg/cleaner/cleaning.go
mrobson: No longer as relevant with the changes, but used the defer for the tarWriter.

defer func() { err = errors.Join(err, tarWriter.Close()) }()

pkg/cleaner/cleaning.go
mrobson: Makes a lot more sense...

Moved the gzip handling to be centralized in ObfuscateFile and obfuscateTar now handles io.Reader/io.Writer with no compression awareness allowing for potential usage for other compressed tars.

tjungblu:
Cool, thanks for adjusting so quickly!

/approve
/lgtm

PR #121, opened 2026-07-24
NO-JIRA: use upstream jsonschema again

tjungblu:
We created our own fork back then to vendor a few upstream PRs for our schema (e.g. enums). Upstream caught up in the meantime and added more features, including everything we have built.

I also changed the additionalProperties flag, with the old fork and `additionalProperties: true,` the generated `UnmarshalJSON` for `SchemaJsonConfig` did a double unmarshal — first into `map[string]interface{}` (the raw
  variable), then into the typed struct. The raw map was needed because with additionalProperties: true, the generator had to capture unknown keys that don't map to struct fields. It just wasn't
  using them for anything meaningful since the Go struct had no AdditionalProperties `map[string]interface{}` field.

  With the upstream go-jsonschema and additionalProperties: false, the generator knows there won't be extra keys, so it skips the raw map unmarshal entirely — it goes straight to type Plain
  SchemaJsonConfig and unmarshals into that. This is both cleaner and slightly more efficient (one unmarshal instead of two).

  The practical impact:

  1. No behavioral change — this project never used or stored additional properties on the config object, so nothing is lost.
  2. Cleaner generated code — the unnecessary raw map[string]interface{} + double unmarshal is eliminated.
  3. Stricter validation — with false, JSON parsers that validate against the schema will reject configs with unknown keys in the config block, which is actually desirable since unknown keys are
  likely typos.

The remaining changes in the generated schema are cosmetic.


swghosh:
@tjungblu Claude wrote up a few more test cases to scrutinize the schema parsing, in the current implementation (on main branch) we successfully pass all of these.

https://github.com/swghosh/must-gather-clean/commit/ee1e0ca589292211b7eae171182a11da39919820

However, one of these failed with the bump:

```
=== RUN   TestSchemaJsonConfigRequiresObfuscateMinItems
    schema_test.go:115: 
                Error Trace:    /Users/swghosh/go/src/github.com/openshift/must-gather-clean/pkg/schema/schema_test.go:115
                Error:          An error is expected but got nil.
                Test:           TestSchemaJsonConfigRequiresObfuscateMinItems
--- FAIL: TestSchemaJsonConfigRequiresObfuscateMinItems (0.00s)
```

i.e. `{"config": {}}` raises an error in the past implementation but with your change it does not.

I'm not sure if this is a blocking issue, just wanted to flag this behaviour!

tjungblu:
Thanks for the additional legwork @swghosh - I believe this is due to the additional properties setting? I'll run this through a debugger in a bit.

Do you mind if I steal your test cases?

swghosh:
> Do you mind if I steal your test cases?

Please @tjungblu :)

tjungblu:
@sairameshv would you have some time to review anytime soon? Sorry for the ping :/

pkg/schema/schema.json
```diff
@@ -114,7 +114,7 @@
                     "description": "RandSeed is the seed to use for priming randomly generated values. When empty or zero, the seed is time.Now().UnixNano(), when set it is honored. It is useful to set for predictable names. It is useful not to set when you want variance in randomly generated names instead of counters to avoid confusion between bugs."
                 }
             },
-            "additionalProperties": true
+            "additionalProperties": false
```
sairameshv, line 117: Changes look good to me, The only suggestion that I have is to add a line in the PR description (or wherever it makes sense), noting the `additionalProperties: true -> false` change

go.mod
```diff
@@ -1,28 +1,28 @@
...
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/sanity-io/litter v1.1.0 // indirect
-	github.com/spf13/pflag v1.0.5 // indirect
-	gopkg.in/yaml.v2 v2.2.2 // indirect
+	github.com/sanity-io/litter v1.5.8 // indirect
+	github.com/sosodev/duration v1.4.0 // indirect
+	github.com/spf13/pflag v1.0.10 // indirect
 )
-
-replace github.com/atombender/go-jsonschema => github.com/tjungblu/go-jsonschema v0.9.1-0.20210922142453-a1b781d84980
```
sairameshv, line 28: /lgtm
/hold

Please feel free to unhold once the above minor nit is addressed.
Thanks

pkg/schema/schema.json
tjungblu: good catch, I've added an explanation about this in the PR description :) Thanks for calling that out.

PR #123, opened 2026-09-07
Add platform auto-detection to skip unnecessary Azure obfuscation

paladugu-anand:
## Problem

Azure resource obfuscation requires a prescan phase that traverses the entire must-gather (reading all files) to discover Azure resource names before the actual cleaning phase. This prescan runs on ALL must-gathers regardless of platform, causing:

- Unnecessary I/O on non-Azure clusters (AWS, GCP, VSphere, BareMetal)
- 15-20% longer processing time for 90-95% of must-gathers
- Confusing warning messages about Azure resources on non-Azure clusters

## Solution

- Reads `cluster-scoped-resources/config.openshift.io/infrastructures/cluster.yaml`
- Extracts `status.platform` field
- Automatically skips Azure obfuscation prescan for non-Azure platforms
- Uses `NoopObfuscator` to maintain obfuscator list consistency
- Adds `--skip-azure` flag for manual override

## Performance Impact

Tested on 8.6GB VSphere must-gather:
- **Before:** 25 minutes
- **After:** 21 minutes
- **Improvement:** 16% faster

## Platforms Affected

- Benefits 90-95% of clusters automatically (non-Azure platforms)
- No change for Azure clusters (prescan still runs)
- Zero breaking changes
- Falls back to running Azure obfuscation if detection fails

## Testing

- All tests pass (`make test`)
- All linting passes (`make verify`)
- Tested on production 8.6GB VSphere must-gather
- Updated test coverage for new `skipAzure` parameter

## Backward Compatibility

- No breaking changes
- Default behavior unchanged for Azure clusters
- All existing config files work without modification
- Report format unchanged

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ pkg/cli/testfiles/aws.cleaned
 test/files/testrun-*
 
 /.claude/
+must-gather-clean-custom
--- a/README.md
+++ b/README.md
@@ -78,6 +78,18 @@ The configuration passed via `-c` is explained in the below [Configuration](#con
 
 By default, the tool runs using multiple threads and is designed to utilize the whole CPU. The number of threads can be adjusted any time with the `-w` argument, defaulting to the number of CPU cores available on the host.
 
+## Platform-Specific Optimizations
+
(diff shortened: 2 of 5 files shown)


tjungblu:
> Zero breaking changes

that's true, but I would appreciate if we do not add anything "platform" specific here. This tool is also used frequently with sos-reports. I can also see that we may need to pre-scan other types of obfuscation providers in the future and then messing with existing skip options becomes very hairy. 

I'll close this if you don't mind, we can maybe revisit this at some point in the future.

/close

